PR #21, opened 2024-06-09
feat: add io and cli

### LINKED ISSUE #4: Feature: command line interface
use case:

```sh
ndv view file.ext
```

ideally, will support as many file types as possible, via some set of plugins... to be determined how to install

tlambert03:
closes #4 

this PR adds a cli command `ndv` ... and attempts to read basically anything :joy: as per @alisterburt request :)

it's a work in progress... would be installed via `pip install ndv[io]` and would lean heavily on aicsimagieo (or bioio, not sure yet) as well as some additional stuff

### DIFF
--- a/src/ndv/io.py
+++ b/src/ndv/io.py
@@ -1,0 +1,43 @@
+"""All the io we can think of."""
+
+from __future__ import annotations
+
+from textwrap import indent, wrap
+from typing import TYPE_CHECKING, Any
+
+import numpy as np
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+    import xarray as xr
+
+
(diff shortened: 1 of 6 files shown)


src/ndv/io.py
```diff
@@ -0,0 +1,43 @@
...
+import numpy as np
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+    import xarray as xr
+
+
+def imread(path: str | Path) -> Any:
+    """Just read the thing already."""
```
alisterburt, line 17: 😂

alisterburt (approved):
Loooove it, just got off the plane but excited to give it a whirl! Thanks for the effort 🙂

PR #22, opened 2024-06-10
feat: chunked loading

tlambert03:
this draft PR moves much of the logic for indexing into data into a Chunker class that takes care of the asynchronous requests to the data store, and yields chunks back to the viewer, which then need only worry about putting them at the appropriate place/channel in the canvas.  It's working pretty well.  

This demo shows:
- we can load data in chunks, regardless of whether the data store is a chunked storage. here, I break up the XY plane into chunks of `(64, 32)`, to show that square chunks are not necessary.  Obviously, the performance benefit here would depend on aligning chunk requests with the chunks of the storage.
- we can specify the order in which to request chunks.  here i load based on the euclidean distance from the center of the array... but it could easily be proximity to any coordinate (such as where the viewer is centered), and could also be used for frustum culling
- when we move the slider, previous requests are canceled, and only the last request goes through
- multi-channel works fine
- (at 0:15) we can still request more dimensions that are being displayed, with the slice slider, and each chunk is responsible for reducing itself down to the number of dimensions being viewed.

https://github.com/pyapp-kit/ndv/assets/1609449/fe4f1980-5373-42bc-a2b6-3b0858c14325

lots of cleanup needed, and need to make sure that it doesn't affect performance of the simple cases

### DIFF
--- a/src/ndv/_chunking.py
+++ b/src/ndv/_chunking.py
@@ -1,7 +1,9 @@
 from concurrent.futures import Future, ThreadPoolExecutor
+from functools import partial
 from itertools import product
 from types import EllipsisType
 from typing import (
+    Callable,
     Deque,
     Hashable,
     Iterable,
@@ -52,7 +54,7 @@ class Slicer:
     ) -> None:
         self.chunks = chunks
         self.data_wrapper: DataWrapper | None = data_wrapper
-        self.executor = ThreadPoolExecutor()
+        self.executor = ThreadPoolExecutor(max_workers=4)
         self.pending_futures: Deque[Future[Response]] = Deque()
 
     def __del__(self) -> None:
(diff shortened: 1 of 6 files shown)


tlambert03:
asynchronous 3d chunked loading!

https://github.com/pyapp-kit/ndv/assets/1609449/c6d4121e-4cd8-423b-be8c-e8973ff256ed

tlambert03:
note to self... some array types (such as tensorstore and dask arrays) provide their own async chunk retrieval and futures patterns.  There should be a way to defer to those for array types that support it

alisterburt:
@tlambert03 finally back from a workshop and checking this out properly - it's wonderful, you're my favorite :)

tlambert03:
:) thanks
after some encouraging proofs of principle, i've gone back to thinking harder about the request/response model.  it's all getting a little blurrier in my head.  I definitely think there's a nice generalizable slicing abstraction to be had here, that doesn't overcomplicate things.   would love to chat about it

alisterburt:
this sounds like fun, let's find a time to chat - will ping

tlambert03:
closing this.  the branch may be useful for future 3d chunking experiments, but the codebase has changed too much in the meantime

PR #23, opened 2024-06-11
feat: Add ROI selection tool

gselzer:
This PR adds tooling for drawing (rectangular) ROIs onto the `NDViewer`. There are still many things left to do (list follows below), but I'm writing this PR to foster discussion and revision.

https://github.com/pyapp-kit/ndv/assets/29754838/1e5439c1-b3cb-4e54-ab51-dfd343b94f04

### Backend
* [x] Write a pygfx implementation

### Features
* [ ] Color widget to alter color as needed for the image. But there's a lot of unknowns here. @tlambert03 is there any color widget in any of the pyapp-kit repos?
* [ ] Different ROI types (ellipse, polyline, polygon, ...) might be nice.
* [ ] Provide API to restrict to a certain maximum number of ROIs. For example, you might only want the user to draw one ROI, and for the old ROI to be deleted when a new ROI is drawn.
* [ ] For the usages, it would be nice for the viewer to expose an event that fires whenever a ROI changes, if that ROI is being used for processing somehow. We likely want a similar event for when the data is edited.

### Bugs
* [x] Currently, it is *very* hard for me to select the ROI, because the ROI can only be selected when you're right on the border. Notably, this only happens when the rectangle is fully transparent. We should figure out whether VisPy provides any aid to make selection easier (there *is* `canvas.visuals_at`, but I haven't investigated. I'm not sure how it handles order, either)
* [ ] Editing the ROI shows a low FPS, significantly lower than when the camera is moved around. That may just be solved by the next point, and maybe also isn't worth worrying about right now...
* [x] Write tests
* [ ] Clean up code, format, resolve TODOs


tlambert03 (commented):
Super clear ❤️.  I like where this is going

tlambert03:
writing future wishes as I have them here.  don't need to be in this PR:

- [ ] option to keep ROIs aligned to data grid (like fiji behavior, to make it very clear which pixels are in the selection)

src/ndv/viewer/_backends/_vispy.py
```diff
@@ -219,3 +487,82 @@ def _on_mouse_move(self, event: SceneMouseEvent) -> None:
...
+                self._selected_roi.select(False)
+            if isinstance(selected.parent, EditRectVisual) or isinstance(
+                selected.parent.parent, EditRectVisual
+            ):
+                self._selected_roi = selected.parent
+                self._selected_roi.select(True, obj=selected)
+                self._selected_roi.start_move(self._canvas_point(event.pos))
+                self._camera.interactive = False
+                handler = qthrottled(
+                    self._on_mouse_move_selection(self._selected_roi), 60
```
tlambert03, line 522: `60` is a bit too long here (was wondering why it felt choppy to move the roi around), let's use 10 or 20

also, let's try to get around `qthrottled(self._on_mouse_move_selection(self._selected_roi))`.  specifically, that pattern of closing on `self._selected_roi` in the wrapped function can cause cleanup issues (since you're passing that closure into `.connect()`.  Need to think about how to improve that

tlambert03 (commented):
i like where this is going.  The *behavior* is lovely.  

I want to spend a little more time thinking about the implementation.  It does "get the thing done", but it's mixing together some concerns that i would prefer to keep separate (in an ideal world).

As we discussed, I think much of it would be made cleaner (and easier on your end) if `ndv` provided you more existing infrastructure to communicate mouse clicks and canvas coordinates to the backends.  So, if it's ok with you, i'd like to pause and think about that bit, specifically with this PR in mind.

src/ndv/viewer/_backends/_vispy.py
gselzer: Agh, is that why it was so slow?

...do we need to throttle it at all?

tlambert03: not necessarily for this case... i throttle the on_slider_change callback since it's a rather expensive operation and we don't want it to run like crazy.  but this is cheap and can probably be removed altogether

gselzer: Removed in 1a7605e

src/ndv/viewer/_backends/_protocols.py
```diff
@@ -3,29 +3,83 @@
...
-    def cmap(self) -> Any: ...
+    def cmap(self) -> cmap.Colormap: ...
     @cmap.setter
-    def cmap(self, cmap: Any) -> None: ...
-    def remove(self) -> None: ...
+    def cmap(self, cmap: cmap.Colormap) -> None: ...
+
+
+# TODO Need a better name :)
+class PHandleHandle(CanvasElement, Protocol):
```
tlambert03, line 60: :joy:

src/ndv/viewer/_viewer.py
```diff
@@ -151,6 +159,9 @@ def __init__(
             QIconifyIcon("fluent:full-screen-maximize-24-filled"), "", self
         )
         self._set_range_btn.clicked.connect(self._on_set_range_clicked)
+        # button to draw ROIs
```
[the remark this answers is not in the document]
gselzer: I've now added the pygfx backend, so this shouldn't be a problem anymore!

src/ndv/viewer/_backends/_vispy.py
```diff
@@ -190,4 +588,12 @@ def canvas_to_world(
         self, pos_xy: tuple[float, float]
     ) -> tuple[float, float, float]:
         """Map XY canvas position (pixels) to XYZ coordinate in world space."""
-        return self._view.camera.transform.imap(pos_xy)[:3]  # type: ignore [no-any-return]
```
gselzer, line 193: @tlambert03 I'm not really sure what the difference is between `self._view.camera.transform` and `self._view.scene.transform`, however I did notice that in 3D mode (at least before any rotation), the latter aligned with what I'd expect (i.e. hovering over the bottom right corner of the image yields world coordinates of `(645, -451, 30)` from the camera's transform and `(256, 256, 30)` from the scene's)

Do you have thoughts here?

tlambert03:
wow, this is really shaping up nicely!

just pulled and played with it. really feels nice,  and I think I like the new `CanvasElement` protocol a lot. 

Still digging through it now, but could you add docstrings to `_protocols.CanvasElement.start_move` and `move`?  (really, I should have docstrings on all the methods defined in there, making it clear what a backend is supposed to accept and return to satisfy the method).   This thought came while trying to determine exactly what start_move vs move gives me

also: amazing that you have the pygfx backend working too.  I did see a slight inconsistency with the handle box.  In pygfx, the "threshold" for how far your mouse can be from the cursor seems to be in data coordinates rather than canavs coords (or something like that) ... if you zoom in, it gets too easy to select the handle:

https://github.com/pyapp-kit/ndv/assets/1609449/6b835a48-a8ba-44b5-b067-32bf1df43b4a

Also, side-note, they look nicer now in pygfx 0.3.0 👍
Screenshot 2024-06-26 at 11 37 40 AM

src/ndv/viewer/_viewer.py
```diff
@@ -290,6 +318,31 @@ def set_data(
         # update the data info label
         self._data_info_label.setText(self._data_wrapper.summary_info())
 
+    def set_roi(
+        self,
+        vertices: list[tuple[float, float]] | None = None,
```
tlambert03, line 323: would be fun to think of ways to overload this method, so you could also pass something like any indexing expression (even something like `np.s_[1:20, 30:60]`) and get an roi.

also.. I expected the to see something about the number of vertices allowed, either in the type hint or the docs.  I eventually found later in the vispy code:

```python
        if len(vertices) != 4 or any(len(v) != 2 for v in vertices):
            raise Exception("Only 2D rectangles are currently supported")
```

should that be enforced here for now?  or is there an alternate path that does currently support other than 4

tlambert03 (commented):
yes, I'm very excited about how the `_viewer.py` looks here!  

I know i still need to dig into the backend implementations, but how are you feeling about all this?  As far as you're concerned is it just waiting for review?  I want to have this in :)

gselzer:
> just pulled and played with it. really feels nice, and I think I like the new `CanvasElement` protocol a lot.

🥳 

> Still digging through it now, but could you add docstrings to `_protocols.CanvasElement.start_move` and `move`? (really, I should have docstrings on all the methods defined in there, making it clear what a backend is supposed to accept and return to satisfy the method). This thought came while trying to determine exactly what start_move vs move gives me

Added!

> also: amazing that you have the pygfx backend working too. I did see a slight inconsistency with the handle box. In pygfx, the "threshold" for how far your mouse can be from the cursor seems to be in data coordinates rather than canavs coords (or something like that) ... if you zoom in, it gets too easy to select the handle:

Yup, as I mentioned privately there's a `FIXME` [here](https://github.com/pyapp-kit/ndv/pull/23/commits/4a18520630d1ee0967742cd8f7f8714dd3512e47#diff-db7795c263de95f0c69e01f15612ee66157eda875a4f2fb85cb3a94e5efccb56R341) about that. As far as I can tell, pygfx defines Point size by specifying the *diameter* of points *in pixels*. The problem is that the math for determining handle selection assumes canvas position and pixels are similarly scaled, which breaks down upon zooming. It's likely that with smarter math, or with a [working `Renderer.get_pick_info`](https://github.com/pyapp-kit/ndv/pull/23/commits/4a18520630d1ee0967742cd8f7f8714dd3512e47#diff-db7795c263de95f0c69e01f15612ee66157eda875a4f2fb85cb3a94e5efccb56R619), we could fix this.

src/ndv/viewer/_viewer.py
gselzer: > would be fun to think of ways to overload this method, so you could also pass something like any indexing expression (even something like `np.s_[1:20, 30:60]`) and get an roi.

Sure! I can't say I have the experience to know what would be nice here, though...

> also.. I expected the to see something about the number of vertices allowed, either in the type hint or the docs. I eventually found later in the vispy code:
> 
> ```python
>         if len(vertices) != 4 or any(len(v) != 2 for v in vertices):
>             raise Exception("Only 2D rectangles are currently supported")
> ```
> 
> should that be enforced here for now? or is there an alternate path that does currently support other than 4

I personally prefer the check within the backend(s), because I wouldn't be surprised if some backends could support more ROI shapes than others later on. We *could* make the error message more specific, though...

Since we're talking about this method, do you think that the specification is general enough? For example, if we later want elliptical ROIs, how could we use `viewer.set_roi` to accomplish that? Similarly, how might we differentiate between polygons and polylines?

tlambert03 (approved):
going in!  Thank you *so* much for all the work on this @gselzer ... really great stuff.

PR #26, opened 2024-06-14
feat: better mouse events and canvas2world methods

tlambert03:
establishes a pattern (using `installEventFilter`) for intercepting mouse and other events before the backend handles them.  And uses it to do a better version of showing the data value over the cursor

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -141,8 +141,10 @@ exclude_lines = [
     "pass",
 ]
 
+
 [tool.coverage.run]
 source = ["ndv"]
+omit = ["src/ndv/viewer/_backends/_protocols.py"]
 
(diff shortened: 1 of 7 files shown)


tlambert03:
cc @gselzer ... your review would be appreciated

gselzer (commented):
@tlambert03 this all looks great to me!

My only question - is there something tricky about adding `PyGFXViewerCanvas.canvas_to_world` now?

tlambert03:
> is there something tricky about adding PyGFXViewerCanvas.canvas_to_world now? 

only that I couldn't figure it out! :) there *must* be a way to do it... but i spent a while and couldn't figure it out.

Issue #8, opened 2024-06-08
todo: add tests for windows


tlambert03:
skipped windows testing at first due to a minor annoyance.  should be easy, just need to do it

tlambert03:
we're testing windows now

PR #27, opened 2024-06-14
Fix leakage in Windows

hanjinliu:
Maybe related to #8.
I found that `qthrottled` holds the reference to the widget instance (via `method.__self__`?) and this prevented proper cleanups in Windows. This problem can be solved by using an weak reference inside a local function as is done in this PR, but probably should be somehow handled on `superqt` side (although I'm not sure if it's possible).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest]
         python-version: ["3.8", "3.10", "3.11"]
 
   test-array-libs:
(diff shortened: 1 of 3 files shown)


tlambert03:
Hey @hanjinliu, lovely seeing you here!  😀 

I saw this as well and added https://github.com/pyapp-kit/superqt/issues/247 to superqt (which is in the latest release)... but I haven't tested on windows yet!

Are you definitely updated on superqt?

hanjinliu:
Hi @tlambert03, I'm very excited about this package!
Sorry for not carefully checking the update... I was using `superqt=0.6.6` and the latest version did solve the issue with my local windows laptop.

tlambert03:
Great!  Good to hear :)

tlambert03:
Shall we still turn this PR into fixing whatever else is left for windows tests?

hanjinliu:
Other fixes will be completely different from what I did here anyway. I think we can just close this PR and make new one later when the solution is found.

tlambert03:
Ok well thanks for popping in!

PR #28, opened 2024-06-15
test: add time-to-show

tlambert03:
this adds a test that ensures that importing ndv and calling imshow() is snappy


gselzer:
Closing because @tlambert03 told me so.

PR #32, opened 2024-06-16
Fix the height of the info label

hanjinliu:
I found that when a `NDViewer` is used inside a dock widget, slicing an n-D image causes "shaking" of the canvas, probably due to the `QSpinner` telling a wrong size (please see below).

https://github.com/pyapp-kit/ndv/assets/40591297/cee3fb1d-22b7-49bc-9d49-26bce12e86b8

This PR simply fixes the height of the widgets.


tlambert03:
lol,  you're using it inside of napari :joy:.

thanks for this! :)

hanjinliu:
No I found this bug elsewhere but napari was the easiest way to add a dock widget 😄

PR #34, opened 2024-06-20
Clean up PyGFX events

gselzer:
This PR adds minor improvements to the PyGFX backend:
* Fixes the event filter, as previously events were not being captured, leading to the hover text not being updated.
* Implements `PyGFXViewerCanvas.canvas_to_world` (inspiration taken from pygfx/pygfx#745). Notably, there *seemed* to be a difference in assumptions between pixel centers between vispy and pygfx - we should decide whether we like the decision I made there.

### DIFF
--- a/src/ndv/viewer/_backends/_pygfx.py
+++ b/src/ndv/viewer/_backends/_pygfx.py
@@ -69,29 +69,17 @@ class PyGFXImageHandle:
 class _QWgpuCanvas(QWgpuCanvas):
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
-        # Every event we want to intercept in the viewer
-        # Must be "ignored" here
-        self._subwidget.mousePressEvent = self._run_and_ignore(
-            self._subwidget.mousePressEvent
-        )
-        self._subwidget.mouseMoveEvent = self._run_and_ignore(
-            self._subwidget.mouseMoveEvent
-        )
-        self._subwidget.mouseReleaseEvent = self._run_and_ignore(
-            self._subwidget.mouseReleaseEvent
-        )
+        self._sup_mouse_event = self._subwidget._mouse_event
+        self._subwidget._mouse_event = self._mouse_event
 
     def sizeHint(self) -> QSize:
         return QSize(512, 512)
 
-    def _run_and_ignore(self, old: Callable) -> Callable:
-        def new(event: QEvent) -> None:
-            # Process the event like normal
-            old(event)
-            # Then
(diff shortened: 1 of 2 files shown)


src/ndv/viewer/_backends/_pygfx.py
```diff
@@ -214,5 +239,31 @@ def canvas_to_world(
         self, pos_xy: tuple[float, float]
     ) -> tuple[float, float, float]:
         """Map XY canvas position (pixels) to XYZ coordinate in world space."""
-        # TODO
-        return (-1, -1, -1)
+        viewport = pygfx.Viewport.from_viewport_or_renderer(self._renderer)
```
gselzer, line 242: This was more or less directly copied from pygfx/pygfx#745, and works nicely, we may want to attribute them though

src/ndv/viewer/_backends/_pygfx.py
```diff
@@ -65,9 +67,32 @@ def remove(self) -> None:
...
 
+    def _run_and_ignore(self, old: Callable) -> Callable:
+        def new(event: QEvent) -> None:
+            # Process the event like normal
+            old(event)
+            # Then
+            event.ignore()
+
+        return new
+
```
tlambert03, line 95:
```suggestion
        self._sup_mouse_event = self._subwidget._mouse_event
        self._subwidget._mouse_event = self._mouse_event

    def sizeHint(self) -> QSize:
        return QSize(512, 512)

    def _mouse_event(
        self, event_type: str, event: QEvent, *args: Any, **kwargs: Any
    ) -> None:
        self._sup_mouse_event(event_type, event, *args, **kwargs)
        event.ignore()
```

src/ndv/viewer/_backends/_pygfx.py
tlambert03: awesome!  yes by all means, add relevant links to the code here.  is this basically the code from `pygfx/helpers/_selection.py`?  If so just mention that file here

tlambert03 (approved):
fantastic!  LGTM

i'll let you add an attribution if you want, and then we can merge

src/ndv/viewer/_backends/_pygfx.py
gselzer: Added

gselzer:
@tlambert03 should be good to go!

src/ndv/viewer/_backends/_pygfx.py
gselzer: Good suggestion!

tlambert03 (approved):

PR #40, opened 2024-08-23
fix: fix canvas update on set_data

### LINKED ISSUE #39: NDViewer doesn't redraw canvas on successive setData calls
* ndv version: 0.0.4
* Python version: 3.12.3
* Operating System: Windows 11

### Description

I've been trying to call `NDViewer.setData()` a couple times, with new data. It turns out that I can call `setData` once and the viewer correctly redraws the data, however after that first time the canvas does not update.

From a bit of digging, I've narrowed the bug down to somewhere in `ndv.viewer._dims_slider.py`. The first time `NDViewer.set_current_index` is called from within `NDViewer.set_data`, it causes a redraw [here](https://github.com/pyapp-kit/ndv/blob/b79241c146bcbbb8160925f8cafbe8734ff75cef/src/ndv/viewer/_dims_slider.py#L409) with `self.value() == {0: 256, 1: 256}`. The **second** time it is called, `self.value() == {}`, and the `self.valueChanged` signal is not emitted.

I further have found that, on both calls to `NDViewer.set_data`, `DimsSlider.setValue` is invoked [here](https://github.com/pyapp-kit/ndv/blob/b79241c146bcbbb8160925f8cafbe8734ff75cef/src/ndv/viewer/_dims_slider.py#L277). However, **only on the first invocation** is [`DimsSlider._on_int_value_changed`](https://github.com/pyapp-kit/ndv/blob/b79241c146bcbbb8160925f8cafbe8734ff75cef/src/ndv/viewer/_dims_slider.py#L345) invoked, leading to the dimension being set within the `DimsSliders.value()` on the first `setData` invocation. This is currently as far as I've gotten.

### What I Did

Here's an MCVE. If you click the "Add image" button once, then you'll get an image of all ones. Further clicks to the button should continue to increment the data, however the canvas doesn't show it:
```python
from qtpy.QtCore import QCoreApplication
from qtpy.QtWidgets import QPushButton, QApplication
from ndv import NDViewer
import numpy
import sys


class Example(NDViewer):

    def __init__(self) -> None:
        super().__init__()
        self.save_btn = QPushButton("Add image")
        self.save_btn.clicked.connect(self.add_image)
        self._btns.insertWidget(0, self.save_btn)
        self.i = 0
    
    def add_image(self) -> None:
        data = numpy.ndarray((256, 256))
        data[:, :] = self.i
        self.i = self.i + 1
        self.set_data(data)


## Most of the code below is copied from ndv.util

def _get_app() -> tuple[QCoreApplication, bool]:
    is_ipython = False
    if (app := QApplication.instance()) is None:
        app = QApplication([])
        app.setApplicationName("ndv")
    elif (ipy := sys.modules.get("IPython")) and (shell := ipy.get_ipython()):
        is_ipython = str(shell.active_eventloop).startswith("qt")

    return app, not is_ipython

app, should_exec = _get_app()
viewer = Example()
viewer.show()
viewer.raise_()
if should_exec:
    app.exec()

```

tlambert03:
fixes #39  
cc @gselzer

### DIFF
--- a/src/ndv/viewer/_lut_control.py
+++ b/src/ndv/viewer/_lut_control.py
@@ -39,6 +39,7 @@ class LutControl(QWidget):
         handles: Iterable[PImageHandle] = (),
         parent: QWidget | None = None,
         cmaplist: Iterable[Any] = (),
+        auto_clim: bool = True,
     ) -> None:
         super().__init__(parent)
         self._handles = handles
@@ -67,7 +68,7 @@ class LutControl(QWidget):
         self._auto_clim = QPushButton("Auto")
         self._auto_clim.setMaximumWidth(42)
         self._auto_clim.setCheckable(True)
-        self._auto_clim.setChecked(True)
+        self._auto_clim.setChecked(auto_clim)
         self._auto_clim.toggled.connect(self.update_autoscale)
 
         layout = QHBoxLayout(self)
@@ -78,6 +79,20 @@ class LutControl(QWidget):
         layout.addWidget(self._auto_clim)
 
         self.update_autoscale()
+
+    def _get_state(self) -> dict[str, Any]:
+        return {
+            "visible": self._visible.isChecked(),
+            "cmap": self._cmap.currentColormap(),
+            "clims": self._clims.value(),
+            "auto_clim": self._auto_clim.isChecked(),
+        }
+
+    def _set_state(self, state: dict[str, Any]) -> None:
+        self._visible.setChecked(state["visible"])
+        self._cmap.setCurrentColormap(state["cmap"])
+        self._clims.setValue(state["clims"])
+        self._auto_clim.setChecked(state["auto_clim"])
 
     def autoscaleChecked(self) -> bool:
         return cast("bool", self._auto_clim.isChecked())
@@ -112,12 +127,11 @@ class LutControl(QWidget):
             clims[1] = max(clims[1], np.nanmax(handle.data))
 
         mi, ma = tuple(int(x) for x in clims)
-        if mi != ma:
-            for handle in self._handles:
-                handle.clim = (mi, ma)
+        for handle in self._handles:
+            handle.clim = (mi, ma)
 
-            # set the slider values to the new clims
-            with signals_blocked(self._clims):
-                self._clims.setMinimum(min(mi, self._clims.minimum()))
-                self._clims.setMaximum(max(ma, self._clims.maximum()))
-                self._clims.setValue((mi, ma))
+        # set the slider values to the new clims
+        with signals_blocked(self._clims):
+            self._clims.setMinimum(min(mi, self._clims.minimum()))
+            self._clims.setMaximum(max(ma, self._clims.maximum()))
+            self._clims.setValue((mi, ma))
--- a/src/ndv/viewer/_viewer.py
+++ b/src/ndv/viewer/_viewer.py
@@ -126,6 +126,7 @@ class NDViewer(QWidget):
         self._img_handles: defaultdict[ImgKey, list[PImageHandle]] = defaultdict(list)
         # mapping of same keys to the LutControl objects control image display props
         self._lut_ctrls: dict[ImgKey, LutControl] = {}
+        self._lut_ctrl_state: dict[ImgKey, dict] = {}
         # the set of dimensions we are currently visualizing (e.g. XY)
         # this is used to control which dimensions have sliders and the behavior
         # of isel when selecting data from the datastore
@@ -300,6 +301,7 @@ class NDViewer(QWidget):
         self.set_current_index(idx)
         # update the data info label
         self._data_info_label.setText(self._data_wrapper.summary_info())
+        self.refresh()
 
     def set_roi(
         self,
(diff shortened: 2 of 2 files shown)


gselzer (commented):
By and large, this looks great, and fixes the behavior observed in #39. Thank you!

The only thing I'll note here is that `NDViewer._update_data_for_index` will be called twice on the first invocation of `NDViewer.set_data`, once explicitly via `NDViewer.refresh` and once through the emission of the `DimsSliders.valueChanged` signal. It's not clear to me how we'd fix that, though, nor whether it's worth fixing...

tlambert03:
fixed, thanks

gselzer:
Aha, didn't realize it could be fixed simply using `signals_blocked`. Good to know!

gselzer:
@tlambert03 I've been experimenting with this branch within my UI, and I've come back to report that `NDViewer.refresh()` has a prohibitive performance impact. I'm not yet sure why, but I'm guessing it's due to slow event handling as performance seems to get worse over time. Here's an MCVE you can try:

```python
from qtpy.QtCore import QCoreApplication, Qt, QTimerEvent
from qtpy.QtWidgets import QApplication
from ndv import NDViewer
import numpy
import sys

# Generate 10 random frames beforehand
frames = [
    numpy.random.randint(0, 255, [256, 256]) for _ in range(10)
]

class ExampleBad(NDViewer):
    """Calls set_data on each timerEvent"""

    def __init__(self) -> None:
        super().__init__()
        self.i = 0
        # Timer triggering frame change every 10 ms
        self._live_timer_id = self.startTimer(
            10, Qt.TimerType.PreciseTimer
        )
    
    def timerEvent(self, a0: QTimerEvent) -> None:
        self.i = (self.i + 1) % len(frames)
        self.set_data(frames[self.i])

class ExampleAlsoBad(NDViewer):
    """Overwrites buffer + calls refresh()` on each timerEvent"""

    def __init__(self) -> None:
        super().__init__()
        # Generate 10 random frames beforehand
        frames = [
            numpy.random.randint(0, 255, [256, 256]) for _ in range(10)
        ]
        self.buffer = numpy.zeros((256, 256))
        self.i = 0
        # Timer triggering frame change every 10 ms
        self._live_timer_id = self.startTimer(
            10, Qt.TimerType.PreciseTimer
        )
        self.set_data(self.buffer)
    
    def timerEvent(self, a0: QTimerEvent) -> None:
        self.i = (self.i + 1) % len(frames)
        self.buffer[:, :] = frames[self.i]
        self.refresh()

class ExampleGood(NDViewer):
    """Overwrites buffer + calls set_current_index()` on each timerEvent"""

    def __init__(self) -> None:
        super().__init__()
        # Generate 10 random frames beforehand
        frames = [
            numpy.random.randint(0, 255, [256, 256]) for _ in range(10)
        ]
        self.buffer = numpy.zeros((256, 256))
        self.i = 0
        # Timer triggering frame change every 10 ms
        self._live_timer_id = self.startTimer(
            10, Qt.TimerType.PreciseTimer
        )
        self.set_data(self.buffer)
    
    def timerEvent(self, a0: QTimerEvent) -> None:
        self.i = (self.i + 1) % len(frames)
        self.buffer[:, :] = frames[self.i]
        self.set_current_index()


## Most of the code below is copied from ndv.util

def _get_app() -> tuple[QCoreApplication, bool]:
    is_ipython = False
    if (app := QApplication.instance()) is None:
        app = QApplication([])
        app.setApplicationName("ndv")
    elif (ipy := sys.modules.get("IPython")) and (shell := ipy.get_ipython()):
        is_ipython = str(shell.active_eventloop).startswith("qt")

    return app, not is_ipython

app, should_exec = _get_app()
viewer = ExampleGood()
viewer.show()
viewer.raise_()
if should_exec:
    app.exec()

```

tlambert03:
yes, this is a very heavy handed refresh approach. I do not really expect `set_data` to be called that often.  perhaps your MCVE is too refined? :). tell me a bit more about the use case that brought you to this.  calling set_data on every data event is not particularly what I imagined we'd do here

gselzer:
> perhaps your MCVE is too refined? :). tell me a bit more about the use case that brought you to this. calling set_data on every data event is not particularly what I imagined we'd do here

My use case is handling a live view from pymmcore-plus (currently using the demo camera). Currently, I've subclassed the `NDViewer` so to override [NDViewer.setData](https://github.com/gselzer/pymmcore-plus-sandbox/blob/da1f7623f50abb8fbf82586f4266d5bbc8fa8ed3/src/pymmcore_plus_sandbox/_viewfinder.py#L193) to be closer to the good example (there may be some excess cruft that should be trimmed off). In an ideal world I wouldn't have to override it, though.

tlambert03:
Not that ndv shouldnt also be able to do that... but have you tried https://pymmcore-plus.github.io/pymmcore-widgets/widgets/ImagePreview/

Thats what we use for rapid image preview

gselzer:
> Not that ndv shouldnt also be able to do that... but have you tried https://pymmcore-plus.github.io/pymmcore-widgets/widgets/ImagePreview/

I do believe that I originally used it, however there are some ndv features (e.g. ROIs, eventually histograms) that @marktsuchida and I are excited to make use of in live views.

tlambert03:
Yep!  Totally understand/agree

PR #41, opened 2024-08-29
feat: RGB support

### LINKED ISSUE #20: todo: support rgb

gselzer:
This PR aims to enable the viewing of RGB datasets. Users specify that their provided data is RGB using the new `ChannelMode` value. The current mechanism enables this mode when the channel axis (specified through existing means) has exactly three indices.

To enable the selection of different modes, the `ChannelModeButton` is refactored into a `ChannelModeComboBox`, which can be altered based on the data to contain the `rgb` mode.

image

This PR requires review, and likely additional minor cleanup.

Closes #20

### DIFF
--- a/src/ndv/util.py
+++ b/src/ndv/util.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import sys
-from typing import TYPE_CHECKING, Any, Literal
+from typing import TYPE_CHECKING, Any
 
 from qtpy.QtWidgets import QApplication
 
@@ -12,14 +12,16 @@ from .viewer._viewer import NDViewer
 if TYPE_CHECKING:
     from qtpy.QtCore import QCoreApplication
 
+    from .viewer._components import ChannelMode
     from .viewer._data_wrapper import DataWrapper
+    from .viewer._viewer import ChannelModeStr
 
 
 def imshow(
     data: Any | DataWrapper,
     cmap: Any | None = None,
     *,
-    channel_mode: Literal["mono", "composite", "auto"] = "auto",
+    channel_mode: ChannelModeStr | ChannelMode = "auto",
 ) -> NDViewer:
     """Display an array or DataWrapper in a new NDViewer window.
 
@@ -29,7 +31,7 @@ def imshow(
         The data to be displayed. If not a DataWrapper, it will be wrapped in one.
     cmap : Any | None, optional
         The colormap(s) to use for displaying the data.
-    channel_mode : Literal['mono', 'composite'], optional
+    channel_mode : Literal['mono', 'composite', 'rgb', 'rgba', 'auto'], optional
         The initial mode for displaying the channels. By default "mono" will be
         used unless a cmap is provided, in which case "composite" will be used.
 
@@ -39,12 +41,8 @@ def imshow(
         The viewer window.
     """
     app, should_exec = _get_app()
-    if cmap is not None:
-        channel_mode = "composite"
-        if not isinstance(cmap, (list, tuple)):
-            cmap = [cmap]
-    elif channel_mode == "auto":
-        channel_mode = "mono"
+    if cmap is not None and not isinstance(cmap, (list, tuple)):
+        cmap = [cmap]
     viewer = NDViewer(data, colormaps=cmap, channel_mode=channel_mode)
     viewer.show()
     viewer.raise_()
--- a/src/ndv/viewer/_backends/_pygfx.py
+++ b/src/ndv/viewer/_backends/_pygfx.py
@@ -32,12 +32,23 @@ def _is_inside(bounding_box: np.ndarray, pos: Sequence[float]) -> bool:
     )
 
 
+# TODO Combine with similar function in _vispy.py
+def _coerce_rgb(data: np.ndarray | None) -> np.ndarray | None:
+    if data is not None and data.ndim == 3:
+        # PyGFX expects (A)RGB data to be X, Y, C
+        for i, s in enumerate(data.shape):
+            if s in [3, 4]:
+                return np.moveaxis(data, i, -1)
+    return data
+
+
 class PyGFXImageHandle:
     def __init__(self, image: pygfx.Image | pygfx.Volume, render: Callable) -> None:
         self._image = image
         self._render = render
         self._grid = cast("Texture", image.geometry.grid)
         self._material = cast("ImageBasicMaterial", image.material)
+        self._cmap = cmap.Colormap("gray")
 
     @property
     def data(self) -> np.ndarray:
@@ -45,7 +56,7 @@ class PyGFXImageHandle:
 
     @data.setter
     def data(self, data: np.ndarray) -> None:
-        self._grid.data[:] = data
+        self._grid.data[:] = _coerce_rgb(data)
         self._grid.update_range((0, 0, 0), self._grid.size)
 
     @property
@@ -85,7 +96,9 @@ class PyGFXImageHandle:
     @cmap.setter
     def cmap(self, cmap: cmap.Colormap) -> None:
         self._cmap = cmap
-        self._material.map = cmap.to_pygfx()
+        # RGB image special case
+        if self.data.ndim != 3:
+            self._material.map = cmap.to_pygfx()
         self._render()
 
     def start_move(self, pos: Sequence[float]) -> None:
@@ -409,7 +422,6 @@ class PyGFXViewerCanvas(PCanvas):
     """pygfx-based canvas wrapper."""
 
     def __init__(self) -> None:
-        self._current_shape: tuple[int, ...] = ()
         self._last_state: dict[Literal[2, 3], Any] = {}
 
         self._canvas = _QWgpuCanvas(size=(600, 600))
@@ -470,6 +482,7 @@ class PyGFXViewerCanvas(PCanvas):
         self, data: np.ndarray | None = None, cmap: cmap.Colormap | None = None
     ) -> PyGFXImageHandle:
         """Add a new Image node to the scene."""
+        data = _coerce_rgb(data)
         tex = pygfx.Texture(data, dim=2)
         image = pygfx.Image(
             pygfx.Geometry(grid=tex),
@@ -479,9 +492,7 @@ class PyGFXViewerCanvas(PCanvas):
         self._scene.add(image)
 
         if data is not None:
-            self._current_shape, prev_shape = data.shape, self._current_shape
-            if not prev_shape:
-                self.set_range()
+            self.set_range()
 
         # FIXME: I suspect there are more performant ways to refresh the canvas
         # look into it.
@@ -504,9 +515,7 @@ class PyGFXViewerCanvas(PCanvas):
 
         if data is not None:
             vol.local_position = [-0.5 * i for i in data.shape[::-1]]
-            self._current_shape, prev_shape = data.shape, self._current_shape
-            if len(prev_shape) != 3:
-                self.set_range()
+            self.set_range()
 
         # FIXME: I suspect there are more performant ways to refresh the canvas
         # look into it.
--- a/src/ndv/viewer/_lut_control.py
+++ b/src/ndv/viewer/_lut_control.py
@@ -135,3 +135,20 @@ class LutControl(QWidget):
             self._clims.setMinimum(min(mi, self._clims.minimum()))
             self._clims.setMaximum(max(ma, self._clims.maximum()))
             self._clims.setValue((mi, ma))
+
+
+class RGBAControl(LutControl):
+    def __init__(
+        self,
+        name: str = "",
+        handles: Iterable[PImageHandle] = (),
+        parent: QWidget | None = None,
+        cmaplist: Iterable[Any] = (),
+        auto_clim: bool = True,
+    ) -> None:
+        super().__init__("RGB", handles, parent, cmaplist, auto_clim)
+        self._cmap.setVisible(False)
+
+    def _on_cmap_changed(self, cmap: cmap.Colormap) -> None:
+        # NB: No-op
+        pass
--- a/src/ndv/viewer/_viewer.py
+++ b/src/ndv/viewer/_viewer.py
@@ -299,6 +299,11 @@ class NDViewer(QWidget):
         # store the data
         self._data_wrapper: DataWrapper = DataWrapper.create(data)  # type: ignore
 
+        # update the range of all the sliders to match the sizes we set above
+        maxes = self._data_wrapper.sizes()
+        with signals_blocked(self._dims_sliders):
+            self._dims_sliders.setMaxima({k: v - 1 for k, v in maxes.items()})
+
         # update the dimensions we are visualizing
         sizes = dict(self._data_wrapper.sizes().items())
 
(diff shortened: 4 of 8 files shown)


src/ndv/viewer/_viewer.py
```diff
@@ -531,17 +547,24 @@ def _update_canvas_data(self, data: np.ndarray, index: Indices) -> None:
...
-            if datum.ndim == 2:
+            if datum.ndim == 2 or self._channel_mode == ChannelMode.RGB:
                 handles.append(self._canvas.add_image(datum, cmap=cm))
             elif datum.ndim == 3:
                 handles.append(self._canvas.add_volume(datum, cmap=cm))
-            if imkey not in self._lut_ctrls:
+            if self._channel_mode == ChannelMode.RGB:
+                # TODO: Is this too strict?
+                for handle in handles:
+                    handle.clim = (0, 256)
```
gselzer, line 566: Maybe this should be labeled as a HACK?

gselzer (commented):
Just wrote down some remaining concerns I had

gselzer:
Converting this back into a draft because additional changes are needed to support smooth transitions between modes

gselzer:
I'm much happier with this after a couple of changes, however I've built it on top of #40, so that PR must be merged first

tlambert03:
merged #40, can you resolve conflicts?

gselzer:
> merged #40, can you resolve conflicts?

Sure!

gselzer:
@tlambert03 conflicts resolved!

tlambert03:
could you add that nifty rgb demo to the examples folder so we have a quick way to test rgb?

src/ndv/util.py
```diff
@@ -29,7 +29,7 @@ def imshow(
         The data to be displayed. If not a DataWrapper, it will be wrapped in one.
     cmap : Any | None, optional
         The colormap(s) to use for displaying the data.
-    channel_mode : Literal['mono', 'composite'], optional
+    channel_mode : Literal['mono', 'composite', 'rgb'], optional
```
tlambert03, line 32: i think `auto` should imply that anything with the last dimension shape as 3 or 4 be rgb.  That is, the example should work with `ndv.imshow(img)` without the transpose and channel_mode specification

src/ndv/viewer/_viewer.py
```diff
@@ -277,26 +277,36 @@ def set_data(
...
-        visualized_dims = list(sizes)[-self._ndims :]
+        sizes = dict(self._data_wrapper.sizes().items())
+
+        visualized_dims = list(sizes)
+        if self._channel_axis is not None:
+            visualized_dims.remove(self._channel_axis)
+        visualized_dims = visualized_dims[-self._ndims :]
         self.set_visualized_dims(visualized_dims)
 
+        is_rgb = (self._channel_axis is not None) and (sizes[self._channel_axis] == 3)
```
tlambert03, line 299: we also need to allow a size of 4 for rgba

src/ndv/viewer/_lut_control.py
```diff
@@ -135,3 +135,20 @@ def update_autoscale(self) -> None:
...
+class RGBControl(LutControl):
+    def __init__(
+        self,
+        name: str = "",
+        handles: Iterable[PImageHandle] = (),
+        parent: QWidget | None = None,
+        cmaplist: Iterable[Any] = (),
+        auto_clim: bool = True,
+    ) -> None:
+        LutControl.__init__(self, "RGB", handles, parent, cmaplist, auto_clim)
```
tlambert03, line 149: just curious, reason not to use `super()` here?  personal preference?

src/ndv/viewer/_viewer.py
tlambert03: also, just curious to discuss what we intend to imply by having "rgb" available here in the combo box.  I guess it means "we could show this to you as rgb if you want"... and perhaps that's ok.  But let's briefly discuss the case of something like a three-channel fluorescence image `(3, Y, X)`.  It's true that we could show you that as RGB, but chances are it won't look reasonable if it's not an 8bit file (since channel scaling would be way off), whereas something that comes in as `(Y, X, 3)` does imply something different about the input.  In either case, it's harmless to show the rgb option in the dropdown, but it may not be _useful_.  worth discussing

tlambert03 (commented):
I'm happy about the direction here!  couple small discussion points.  Mostly, we need to do a little magic/heuristics here and i think it's worth laying out exactly what assumptions we are going to make about data formats coming in

src/ndv/viewer/_lut_control.py
gselzer: Oops, that's likely a remnant of earlier experimentation. Fixed!

src/ndv/viewer/_viewer.py
gselzer: > also, just curious to discuss what we intend to imply by having "rgb" available here in the combo box.

Yeah, the thought was that a 3-channel image may or may not be RGB, and it would be wrong to assume such. If the user does not specify the mode, then 3-channel images still appear with `ChannelMode.MONO`. But then if you *want* to view in RGB, you can do so, as the option is now available in the combo box.

> But let's briefly discuss the case of something like a three-channel fluorescence image `(3, Y, X)`. It's true that we could show you that as RGB, but chances are it won't look reasonable if it's not an 8bit file (since channel scaling would be way off),

Yup, this was the reasoning behind `ChannelMode.MONO` remaining the default. 

> whereas something that comes in as `(Y, X, 3)` does imply something different about the input.

I'm unsure if this was explicitly mentioned, but passing data in *either* `(Y, X, 3)` *or* `(3, Y, X)` should enable RGB viewing i.e. if you have a channel axis (discovered using existing logic) with 3 indices, RGB mode is enabled (and we can add 4 indices if desired).

By and large, we wanted to avoid making assumptions about the data, enabling an RGB mode wherever possible without assuming that the user wants it.

gselzer:
@tlambert03 added a WIP commit fleshing out alpha support. Couple questions:
* Shall we rename the channel mode (e.g. `rgba`?)
* Should we have multiple examples? Do you like the alterations made to `examples/rgb.py` in that commit?

src/ndv/util.py
gselzer: As I mentioned below, users should not have to transpose their data - RGB mode will use the existing logic to seek out the channel axis, and transpose the data as needed.

As far as the channel mode spec goes, if you like the idea of automatically using RGB mode for `(Y, X, 3)` or `(Y, X, 4)` data here I can add it in...

gselzer:
Oh, and one more discussion point @tlambert03:

[These `SETUP` lines](https://github.com/pyapp-kit/ndv/blob/8f9212d91aac9a60e67dddc35d302aec6323fe22/src/ndv/viewer/_viewer.py#L237) have given me a bit of a headache. I've found that if you set the channel mode to RGB before setting the data, the channel axis is present as a slider, until you select a different mode (because `NDViewer.set_channel_mode` contains the logic for removing the channel slider if not mono mode). If you set the channel mode after, we end up in this strange intermediate state where RGB mode is "selected", but I can only see the final channel of the data. Thoughts?

src/ndv/util.py
tlambert03: ahh, true.  In that case (and in any case really) I think it makes sense to move the "auto" option & logic to the viewer `__init__` itself, and not just this convenience function

tlambert03:
> Oh, and one more discussion point @tlambert03:
> 
> [These `SETUP` lines](https://github.com/pyapp-kit/ndv/blob/8f9212d91aac9a60e67dddc35d302aec6323fe22/src/ndv/viewer/_viewer.py#L237) have given me a bit of a headache. 

played around with this a bit... i feel like setting channel mode second "feels" more correct (since it's more likely to be changed in the course of viewing).  does this work for you? https://github.com/gselzer/ndv/pull/3

src/ndv/util.py
gselzer: As of 339607c, it has been moved within `NDViewer`'s construction.

src/ndv/viewer/_viewer.py
gselzer: alpha support has been added with 339607c, and `(Y, X, 3)`/`(Y, X, 4)` images are now displayed as RGB(A) when the channel mode is `auto`. Notably, `(3, Y, X)` images are not.

gselzer:
>     * Shall we rename the channel mode (e.g. `rgba`?)

I took the liberty of renaming it!

tlambert03:
This is all looking good and we can finish it up this week,  I was thinking that the main thing to fix that race condition is probably just to remove the conditional check for existing image handles (there's little harm in calling refresh accidentally, and it will cancel any waiting futures anyway).  So, that, combined with putting the call to set channel mode after set data mode in the init should be good I hope

src/ndv/viewer/_viewer.py
```diff
@@ -425,6 +444,21 @@ def set_current_index(self, index: Indices | None = None) -> None:
...
+        self,
+        colormaps: Iterable[cmap._colormap.ColorStopsLike] | None = None,
+        data: DataWrapper | Any | None = None,
+    ) -> ChannelMode:
+        # Users who provider colormaps generally expect composite images
+        if colormaps is not None:
+            return ChannelMode.COMPOSITE
+        # Data shaped [Y, X, 3] or [Y, X, 4], are usually RGB images
+        if (shape := getattr(data, "shape", None)) is not None:
+            if len(shape) == 3 and shape[-1] in [3, 4]:
```
tlambert03, line 457: i think we could remove `len(shape) == 3`?

src/ndv/viewer/_backends/_pygfx.py
```diff
@@ -470,6 +479,12 @@ def add_image(
         self, data: np.ndarray | None = None, cmap: cmap.Colormap | None = None
     ) -> PyGFXImageHandle:
         """Add a new Image node to the scene."""
+        if data is not None and data.ndim == 3:
+            # PyGFX expects (A)RGB data to be X, Y, C
+            for i, s in enumerate(data.shape):
+                if s in [3, 4]:
+                    data = np.moveaxis(data, i, -1)
+                    break
```
tlambert03, line 486: share this in a helper function somewhere

src/ndv/util.py
```diff
@@ -19,7 +19,7 @@ def imshow(
     data: Any | DataWrapper,
     cmap: Any | None = None,
     *,
-    channel_mode: Literal["mono", "composite", "auto"] = "auto",
+    channel_mode: Literal["mono", "composite", "rgba", "auto"] = "auto",
```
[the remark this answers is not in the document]
gselzer: Fixed with 8fcb5ba

src/ndv/viewer/_viewer.py
gselzer: [e4755a5](https://github.com/pyapp-kit/ndv/pull/41/commits/e4755a5866f029853abe93392718608170d40758)

src/ndv/viewer/_backends/_pygfx.py
```diff
@@ -85,7 +92,9 @@ def cmap(self) -> cmap.Colormap:
     @cmap.setter
     def cmap(self, cmap: cmap.Colormap) -> None:
         self._cmap = cmap
-        self._material.map = cmap.to_pygfx()
+        # FIXME: RGB image special case
```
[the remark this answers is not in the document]
gselzer: Fixed with [80dae42](https://github.com/pyapp-kit/ndv/pull/41/commits/80dae42cc75851dbaac8c429985fd0f10bb73e32) - shame GitHub was erroring trying to apply your suggestion!

src/ndv/viewer/_backends/_pygfx.py
gselzer: DRYer with [e095b60](https://github.com/pyapp-kit/ndv/pull/41/commits/e095b6043c7649f1103b3a119216f34a63c40b58)

tlambert03:
in [e1ae37c](https://github.com/pyapp-kit/ndv/pull/41/commits/e1ae37cbfb3967fc01b64c39b669ebaf6da8e193) I removed the conditional checks before calling `self.refresh()` ... i'm curious to know whether you can still hit the problem mentioned in https://github.com/pyapp-kit/ndv/pull/41#issuecomment-2356287035

gselzer:
> in [e1ae37c](https://github.com/pyapp-kit/ndv/pull/41/commits/e1ae37cbfb3967fc01b64c39b669ebaf6da8e193) I removed the conditional checks before calling `self.refresh()` ... i'm curious to know whether you can still hit the problem mentioned in [#41 (comment)](https://github.com/pyapp-kit/ndv/pull/41#issuecomment-2356287035)

I cannot see either issue now! No slider, and I'm seeing an rgba image! Great!

tlambert03:
seems that:

```python
# run rgb example then...

In [15]: wdg.set_data(ndv.data.cells3d())

In [16]: wdg.set_data(img)
```

does show img as rgb, but composite is still selected in the combobox.  

(sidenote: cells3d is shown as noise ... which is weird, but probably unrelated here)

src/ndv/viewer/_viewer.py
```diff
@@ -277,26 +281,38 @@ def set_data(
             or slices that define the slice of the data to display.  If not provided,
             the initial index will be set to the middle of the data.
         """
+        if hasattr(self, "_data_wrapper"):
```
tlambert03, line 284: maybe just add a comment for what you think you were avoiding here

src/ndv/viewer/_viewer.py
```diff
@@ -593,6 +644,8 @@ def _reduce_data_for_display(
                 data = data.astype(np.int32)
             else:
                 data = data.astype(np.float32)
+        if self._channel_mode == ChannelMode.RGBA:
+            data = data.astype(np.uint8)
```
tlambert03, line 648: maybe add a warning here before casting?

gselzer: @tlambert03 and I decided to remove this casting to allow for e.g. uint16 data, **[0-1] normalized** floating point data, etc.

We can always restrict this later if requested.

gselzer:
@tlambert03 I'd really like to squash all these commits together soon, as the history is rather messy, and do one final review.

This script is helpful for testing:
```python
import math

import numpy

import ndv
from qtpy.QtWidgets import QApplication, QPushButton

# 2D RGB dataset
rgb = numpy.zeros((256, 256, 4), dtype=numpy.uint8)
for x in range(256):
    for y in range(256):
        rgb[x, y, 0] = x
        rgb[x, y, 1] = y
        rgb[x, y, 2] = 255 - x
        rgb[x, y, 3] = int(math.sqrt((x - 128) ** 2 + (y - 128) ** 2))

# 4D ZCYX dataset
cells = ndv.data.cells3d()

# Button to switch between them
def switch(checked: bool) -> None:
    n.set_data(rgb if checked else cells)
btn = QPushButton("Switch image")
btn.setCheckable(True)
btn.toggled.connect(switch)

# Run it!
app = QApplication([])
n = ndv.NDViewer()
n._btns.addWidget(btn)
n.show()
app.exec()

```

gselzer:
Superseded by #141

Issue #49, opened 2024-10-10
Add histogram


tlambert03:
opening an issue to discuss adding a histogram (which everyone agrees is a must-have feature)

@gselzer has done a lot of work here, currently in https://github.com/gselzer/ndv/tree/interactive-clims

I had a similar vispy-backed histogram that controlled the napari LUTs back in https://github.com/napari/napari/pull/675 ... at one point, I had put a lot of work/consideration into it, and that branch is still at https://github.com/tlambert03/napari/tree/histogram

### backend considerations

we have a few options for tools we could use to create the actual histogram graph

- Qt's [Graphics View](https://doc.qt.io/qt-6/graphicsview.html).  This is what's used, for example, in [pyqtgraph's histogram](https://pyqtgraph.readthedocs.io/en/latest/api_reference/graphicsItems/histogramlutitem.html).  The benefit here is that it would work regardless of what canvas backend we use for the image itself (e.g. vispy or pygfx).  The downside is that it won't translate easily to a future jupyter widget.
- Directly create them in backends like vispy (as @gselzer did in his branch and I did in https://github.com/napari/napari/pull/675).  This has the advantage of being easily portable to jupyter (see #48), but the disadvantage of needing to be implemented in each backend.  (which again, can re-raise the discussion of whether we want to keep supporting multiple canvas backends)

### wants

here's a partial list of things that one might want in a histogram.  not all of these are mandatory for an initial implementation.

- axes with ticks & numbers that pan/zoom with the histogram itself
- full styling API: colors for face, background, text, ticks etc... font sizes, line widths
- ability to turn off stylings for a bare minimal histogram (e.g. that could be put on a slider as in @gselzer's demo)
- separation of histogram calculation logic from display logic
- min/max/gamma controls for connection with a LUT model
- some consideration for how we will display multiple channels (in fluorescence) or overlaid RGB

gselzer:
If you want to assign this to me @tlambert03 I'm happy to keep hacking along 😄 

I had some further discussions with @marktsuchida and @nicost yesterday, which I'll summarize below:

## Additional Wants:
* Toggle-able logarithmic y-axis to view distribution tails
* Enable removal of a variable percentage of outliers in autoscale (ignore percentage in mmstudio)
* Display of additional statistics (mmstudio has average, standard deviation)
* Restrict calculation within ROIs
[...]

Issue #48, opened 2024-10-10
Feature: make useable with Jupyter


tlambert03:
This issue is to track support for jupyter notebook.  As mentioned in the readme, while this package began as a Qt widget, it is within scope of the aims to provide a jupyter front end as well.

Both vispy and pygfx have ways of rendering a canvas to an offscreen/remote frame buffer, which can be rendered in an `img` tag in the browser.  things like `jupyter-rfb` make that possible.

In order to achieve this, a better model/view separation will be required (see # for progress on that)

We also currently make use of superqt for things like range sliders and cmap selection comboboxes... so those would also need to be figured out for jupyter widgets... but anywidget and @mantz are likely to provide answers there :)

tlambert03:
see https://github.com/pyapp-kit/ndv/pull/51#issuecomment-2420152055 for working proof of principle

PR #51, opened 2024-10-12
feat: ArrayDisplayModel (for later model/view refactor)

tlambert03:
this is the model-only partly from #33, cleaned up for easier review.  This isn't yet connected to a view, but can be reviewed for soundness of the model itself, interactive validation (i.e. instantiate the object and see if you can break it with invalid inputs), and signal emission (they should be appropriate for later view connection).

see the `display_model.py` example for a starting point

cc @gselzer if you want to play with it and provide feedback (including requests for more docstrings & comments for any parts that seem confusing or magical), or try to connect it to a view


src/ndv/models/_base_model.py
```diff
@@ -0,0 +1,14 @@
+from typing import ClassVar
+
+from psygnal import SignalGroupDescriptor
+from pydantic import BaseModel, ConfigDict
+
+
+class NDVModel(BaseModel):
+    """Base eventd model for NDV models."""
```
gselzer, line 8: `eventd`

gselzer (commented):
@tlambert03 this is a great start, thanks for making it public so quick!

> reviewed for soundness of the model itself

I think that the model is sound, but am wondering about your plans for modeling the *other* data involved, namely the array data itself (which right now is just a wrapper, which might be fine) which might have associated metadata (like pixel scale, offsets) relevant to the display, as well as ROIs.

It's also worth denoting whether this model could handle multiple datasets (which is a big can of worms we can address later, but it's still worth thinking about for model robustness). I think that, written right now, it could, but is that a "feature" we should preserve later, or just an aspect of the current state it maintains?

>  interactive validation (i.e. instantiate the object and see if you can break it with invalid inputs)

Buggy things like this could be improved later, but fwiw I don't think `m.current_index["5"] = slice(1, "fooo")` should be allowed 😆.

> signal emission

Don't have much to say here, other than that I'd love to try hooking this up to a widget and see how far I/we can get.

tlambert03:
> am wondering about your plans for modeling the other data involved, namely the array data itself

just pushed some of my recent playground changes (in `viewer_v2.py`) where you can see how I was envisioning this.  Basically, the array display model is intentionally expressed independent of the data itself.  But of course you will need to pair it with a DataWrapper in order to get much use.   I'm currently playing with that in the Viewer object itself, but if there's sufficient logic just between that pair of objects, it could be extracted to an intermediate object.

> whether this model could handle multiple datasets. I think that, written right now, it could, but is that a "feature" we should preserve later, or just an aspect of the current state it maintains?

yes that is definitely a goal here (and is part of why i exclude the dataset from the display model).  I don't think it's a big can of worms really... it just requires a more explicit model of a "world object" (i.e. a thing that has a data-to-world transform) that tells us how to place the object in the scene graph.  Essentially, right now you can think of it as everything having an (implicit) data-to-world transform of "identity".  

basically, i think everything's still on a good path here for these things

> I'd love to try hooking this up to a widget and see how far I/we can get.

see x.py for now in this PR

tlambert03:
> m.current_index["5"] = slice(1, "fooo")

the only issue there is the "fooo" in the slice right?  I can update the Slice type validator

gselzer:
> just pushed some of my recent playground changes (in `viewer_v2.py`) where you can see how I was envisioning this. Basically, the array display model is intentionally expressed independent of the data itself. But of course you will need to pair it with a DataWrapper in order to get much use. I'm currently playing with that in the Viewer object itself, but if there's sufficient logic just between that pair of objects, it could be extracted to an intermediate object.

Oh, great!

Since you're already working on the "view" part...how much are you planning to separate the "view" part (Qt/Jupyter, vispy/pygfx) from the "controller" part (signal connections, maybe some top-level user API)? Might be helpful in resolving #48

> yes that is definitely a goal here (and is part of why i exclude the dataset from the display model). I don't think it's a big can of worms really... it just requires a more explicit model of a "world object" (i.e. a thing that has a data-to-world transform) that tells us how to place the object in the scene graph. Essentially, right now you can think of it as everything having an (implicit) data-to-world transform of "identity".
> 
> basically, i think everything's still on a good path here for these things

Glad to hear it! Since we're talking about transforms, do you think it's worth allowing `current_index` values be numbers (i.e. `int` or `float`)?

> see x.py for now in this PR. Note, at this point it's just in "hook all the things up mode" ... not clean and not refactored to be clear

❤️ 

> the only issue there is the "fooo" in the slice right? I can update the Slice type validator

Yeah, think so!

tlambert03:
> how much are you planning to separate the "view" part (Qt/Jupyter, vispy/pygfx) from the "controller" part (signal connections, maybe some top-level user API)? Might be helpful in resolving

was definitely also in scope for this :) i just split some stuff up and now there's an `mvc.py` pattern.  Still a mess, but you can run it.  Gives a decent example of how we can split all front-end concerns

gselzer:
> was definitely also in scope for this :) i just split some stuff up and now there's an `mvc.py` pattern. Still a mess, but you can run it. Gives a decent example of how we can split all front-end concerns

Gorgeous 😍 this was exactly what I was envisioning, and I was hacking out something like this before I saw your WIP commit yesterday.

tlambert03:
something I'm still thinking about a lot (as you were above) is the relationship between the data wrapper and the display model.

I do definitely want it to be possible to have the same display model associated with two different data wrappers (this is essentially like "synced windows" in Fiji, where you link the current index and luts of two different images).  And I also want it to be possible to associate two different display models with the same data wrapper (this would make orthoviewers trivial).  So I don't think I want to simply add a `data` field to the `ArrayDisplayModel`.  But i do find it a little clunky to pass around both the "model" and the "data" (since they are rather linked with each other).  So perhaps an intermediate object:

```python
class DisplayedData:
    data: DataWrapper | None
    view_model: ArrayDisplayModel
```

tlambert03:
> Since we're talking about transforms, do you think it's worth allowing current_index values be numbers (i.e. int or float)?

this concept is important if we want to select a display in world space (instead of data space), but that sort of thing was such a hairy ball of wax in napari, that I want to leave all that kind of multi-world-object negotiation for another PR

gselzer:
> That object would have logic like the [_canonicalize_axis_key](https://github.com/pyapp-kit/ndv/pull/51/files#diff-2285cb09830aaf8b30e404bbd653385b810a44aa97c6131dc746858bfdd95321R138-R152) method. That method is essentially there because we do want someone to be able to express axis keys in the display in as flexible of a way as possible (including negative indexing, named axes, etc... visible_axes = (-2, 1) or visible_axes = {'y', 'x'}). But at the end of the day, those keys must be resolved to a specific axis index, and for that we need the data. So that's a good place for it i think

It's possible that I just haven't dug as deep in as you have yet, but why can that not be resolved in the view, or in the controller? I do like the idea of thinking about the "Viewer" having three things, which it is responsible for interconnecting:
1. `data`: the information being displayed
2. `model`: how to display `data` (this definition, which is basically copied from `_array_display_model`, could use a little wordsmithing to sufficiently distinguish from the `view`)
3. `view`: the mechanism of display

gselzer:
I've also been hacking more on #49 in the meantime, now off of your prototype. Do you think that it makes sense to build up `StatsModel` and `StatsView` objects in tandem with the objects you're building here?

tlambert03:
> Do you think that it makes sense to build up StatsModel and StatsView objects in tandem with the objects you're building here?

I do actually think that model view could work there.  That's a better way to phrase the vague things I was saying about some "stats service" (i.e. model), the thing that can calculate min max mean histogram, etc... _once_, and be shared by various things, like the auto-scaler, the histogram, etc...

gselzer:
Yup, and I also like the idea of a view protocol because it could make it easier to define multiple histogram widgets as I describe in #49 - a simple, small one that could replace dimension sliders, and a more complicated widget like you had in napari/napari#675 that could be brought up with double click...

tlambert03:
proof of principle working with Jupyter :)


https://github.com/user-attachments/assets/041de589-ae2e-400b-bb92-c34776286224

src/ndv/models/_data_display_model.py
```diff
@@ -0,0 +1,153 @@
...
+    def isel(self, index: Mapping[int, int | slice]) -> np.ndarray: ...
+
+
+class DataDisplayModel(NDVModel):
+    """Combination of data and display models.
+
+    Mostly this class exists to resolve AxisKeys in the display model
+    (which can be axis labels, or positive/negative integers) to real/existing
+    positive indices in the data.
+    """
```
gselzer, line 32: I think it's unfortunate that this class is necessary - it starts to tie together one display model to one dataset, which could restrict our ability to display multiple datasets using a single data model. I'd like to look more into whether this is really necessary (I'm guessing it's hairy since you had to make this, but maybe a fresh set of eyes will help?)

src/ndv/controller/_controller.py
```diff
@@ -0,0 +1,180 @@
...
+        if data is None:
+            data = DataDisplayModel()
+        if view is None:
+            view = get_view_backend()
+        self._dd_model = data  # rename me?
+        self._view = view
+
+        self._set_model_connected(self._dd_model.display)
+        self._view.currentIndexChanged.connect(self._on_view_current_index_changed)
+        self._handles: dict[int | None, PImageHandle] = {}
```
gselzer, line 30: Could these handles be a property of the view?

I'm guessing that they're here because the view needs to be passed the data contained in the `DataDisplayModel`, but what if the data was instead a property of the *view*?

src/ndv/models/_data_wrapperold.py
```diff
@@ -151,6 +159,26 @@ def guess_channel_axis(self) -> Hashable | None:
     def save_as_zarr(self, save_loc: str | Path) -> None:
         raise NotImplementedError("save_as_zarr not implemented for this data type.")
 
+    # TODO: we need a way for DataWrapper to be able to say that the dims/coords
+    # have changed (essentially, that the shape of the data has changed)
+    # among other things, this would allow the canonicalized_axis_map to be cleared
```
gselzer, line 164: I think in a lot of ways the `DataWrapper` is in and of itself a Model class - if it was a `DataModel` instead, things could listen to the data changing?

src/ndv/views/_jupyter/jupyter_view.py
```diff
@@ -0,0 +1,168 @@
...
+        self._cmap.value = cmap.name
+
+    def setClims(self, clims: tuple[float, float]) -> None:
+        self._clims.value = clims
+
+    def setLutVisible(self, visible: bool) -> None:
+        self._visible.value = visible
+
+
+class JupyterViewerView:
```
gselzer, line 85: ❤️ 

...also great name

src/ndv/models/_data_display_model.py
tlambert03: Yeah that class is still up in the air for me too, it may be removed...  it exists independently specifically because we do kinda need one place to tie one model (which has abstract axis keys that may or may not mean anything) to one dataset (where those keys are canonicalized into actual dimension indices or errors).  And this class does that.  For example if you had a model that says `visible_dims = ('t', 'x', 'y'), but a dataset comes along that has no t, we need to resolve and check that _somewhere_

src/ndv/controller/_controller.py
tlambert03: They're actually there more because it's vispy/pygfx that provide the handles, and I'd like to be able to use that same logic regardless of the front end (qt/jupyter) view. 

But this is a good thing to flag 👍, this isn't a pattern I'm married to at all

src/ndv/models/_data_wrapperold.py
tlambert03: Yeah, it's model-like for sure.  And yes, it needs some signals.  Namely dimsChanged and perhaps coordsChanged. dataChanged itself is really quite hard (something we struggled with in napari a lot), because it can be very hard to spy on in-place mutations for many array types.  But it would still be good to have a signal there that could at least be manually emitted

tlambert03:
to prevent confusion, I'm going to close this PR.  As I told @gselzer, this branch is now present upstream at https://github.com/pyapp-kit/ndv/tree/v2-mvc ... and to facilitate multiple people (me and gabe) working on different aspects of it at the same time, all PRs working on the model view refactor will be made to that `v2-mvc` branch instead of to `main`.  Then, we will eventually merge that branch into main in one PR

PR #52, opened 2024-10-18
feat: Histogram Widget

gselzer:
This PR adds a standalone widget for creating histograms of datasets. This PR is limited in scope, as requested by @tlambert03 - it does not interface with ndv in any way.

This PR is designed for integration with #51, and adopts a MVC structure to fit with the components being built over there. The components are described below:

## `StatsModel`

This component is designed to retain computed statistics from a provided dataset (right now, it's just the histogram, really, but we may want more later), and signal listeners when the dataset changes.

## `StatsView`

This component contains the API to describe an object that wants to listen for changes to the `StatsModel`, and update some display of that data for observation. One example might be a `QWidget` with some labels displaying mean and standard deviation, like exists in mmstudio.

## `LutView`

This component contains the API to describe an object that wants to listen for changes to a LUT being used to visualize values in a dataset, and to update some display of that data for observation. The current `ndv` `LutControl` objects could be refactored into one of these. I believe that @tlambert03 has a `LutModel` as a part of #51 - we should coordinate here to make sure the API lines up.

## `HistogramView`

This component is both a `StatsView` (because the histogram is a statistical feature of a dataset) and a `LutView` (because it has controls for clims, gamma, etc.), with some extra API pertaining to histogram viewing.

## `VispyHistogramView`

This component is an implementation for `HistogramView`, backed by VisPy, and the real workhorse of the PR. 

If you run `x.py`, you'll see the following, which is a  `VispyHistogramView` that has been connected to a `StatsModel`. Right now, there is no controller, because 🤷 

image

# TODO
...I'm tired, but this will be an active list of wants/needs/dreams...

* [ ] Make sure things align with #51 
* [ ] Resolve TODOs
* [x] Add vertical mode (+ API)
* [ ] Clean & Test (an interactive example that changes the data would be cool)
  * [x] Test log Y
* [ ] PyGFX backend?


tlambert03:
honestly, I think this is pretty great!  I could certainly make tiny little nitpicky comments or suggestions, but they all feel rather bike-sheddy.  I like the design a lot.  There are perhaps a few usability quirks when using various methods like `enable_range_log` and other tiny little things that I came across, but i didn't write any of them down yet since they all felt rather insignificant on the whole.  So, I'm very happy with the direction here.  some open questions in my mind include whether most of this logic could love on some base class (kinda on the HistogramView Protocol object) that would implement anything that vispy or pygfx wouldn't directly need to care about, but i would just keep going with the vispy object for now as you are doing, and then look at refactoring out the backend agnostic stuff later.

since I don't have many concrete suggestions at the moment, you could just keep working on polishing things you know you wanted to work on.  Or I could just keep playing with it and make tiny changes as I go that I could either push directly or in a sub-PR (sometimes that's easier for me than making github comments).

great work 🚀

tlambert03:
in terms of alignment with #51, I'd say don't worry too much.  I don't think that the LUTs signals are going to change all that dramatically; so whoever merges second can just make whatever little patches would be necessary if they diverge slightly

tlambert03:
oh, and by the way, I do think we should look into seeing what fastplotlib can provide for the pygfx side of things.  beyond the package itself, it brings no additional dependencies (beyond stuff we already depend on), so it could simply be added to the `pygfx` extra here

gselzer:
> [ ] when dragging the clims sliders, one shouldn't be able to drag one line "past" the other (i.e. `clims[1]` must be greater than `clims[0]`)

@tlambert03 tell me what you think of 5a7097b.

tlambert03:
I was already wondering whether using composition (rather than directly inheriting sceneCanvas) might work better,  and I think that would maybe solve this too (but may add problems too, need to look closer)

gselzer:
@tlambert03 I've made some significant changes here, including your composition suggestion, and created an interactive example within `x.py` - I'm going to try hacking on a pygfx example (will first look at fastplotlib as per your suggestion), but if you get a chance to look at the example I'd be interested in your thoughts

gselzer:
> oh, and by the way, I do think we should look into seeing what fastplotlib can provide for the pygfx side of things. beyond the package itself, it brings no additional dependencies (beyond stuff we already depend on), so it could simply be added to the `pygfx` extra here

Looking at fastplotlib, the library has a [HistogramLUTTool](https://github.com/fastplotlib/fastplotlib/blob/f0f80c6a4deabb0941154584c854f2004b0cc316/fastplotlib/tools/_histogram_lut.py#L27), which is great, however:
* It doesn't take a precomputed histogram - maybe we could alter the design of the current `StatsModel` to just pass off the data, but I like the idea of precomputing the histogram.
* It's a bit buggy - for example, you can rotate the graphic like I do in that script, but the canvas events don't reflect that, so dragging clims behaves unexpectedly 😅 
* No gamma curve is unfortunate.

Also, it's worth noting that the histogram does not provide the same customization as VisPy, which isn't a dealbreaker but is unfortunate.

I ran out of time to determine whether a logarithmic axis is possible - guessing so, but didn't have time to find it.

All of these points are based off of the latest version on PyPI right now - it's possible that the latest `main` has fixed many of these. 

image


Starter example code below:
```python
# test_example = true

from math import pi
import fastplotlib as fpl
from fastplotlib.widgets.histogram_lut import HistogramLUT
import numpy as np

figure = fpl.Figure(size=(700, 560))
# cmap_transform from an array, so the colors on the sine line will be based on the sine y-values
data = np.random.normal(10, 10, 10000).reshape((100, 100))
ig = fpl.ImageGraphic(data)
h = HistogramLUT(data, ig)
h.rotate(-pi / 2, "z")
g = figure[0, 0].add_graphic(h)
figure.show()


# NOTE: `if __name__ == "__main__"` is NOT how to use fastplotlib interactively
# please see our docs for using fastplotlib interactively in ipython and jupyter
if __name__ == "__main__":
    print(__doc__)
    fpl.run()
```

tlambert03:
ok, that's great detail, thanks for looking into it.  in that case, maybe we're better off vendoring/borrowing little bits more as a pattern of usage than a direct dependency

gselzer:
> maybe we're better off vendoring/borrowing little bits more as a pattern of usage than a direct dependency

I believe that general figure structure is still valuable and maybe worth the dependency, so I'm currently trying to hack together a `Graphic` subclass that can do what we want. Who knows, it may even be worth putting something together that could be contributed upstream?

tlambert03:
just checking in here.  here are things i think should happen before merge (and things that we don't need to do in this PR)

### need

- [ ] the clims sliders shouldn't be allowed to pass each other, as mentioned above.  I'm not quite sure what https://github.com/pyapp-kit/ndv/commit/5a7097bfc7ac71f49ba812d5a81d4051ef1e84e2 is doing behaviorally with `newlims[:] = newlims[::-1]`?  I would expect the slider to just stop and not go any farther (and pin the newlims to the min/max of the other value accordingly)... but it kinda looks like you're auto-switching the grabbed handle to the other limit?
- [ ] need tests.  I'm not worried about testing the visual/on-screen representation for now. but smoke testing all public methods for exceptions should happen.
- [ ] linting and pre-commit
- [ ] open question: I wonder whether `_clim_handle_grabbed` and `_gamma_handle_grabbed` should be combined into a single `grabbed_item` concept, that has an Enum of possible things to grab:
    ```python
    class Control(Enum):
        LEFT_CLIM = auto()
        RIGHT_CLIM = auto()
        GAMMA_HANDLE = auto()
    ```
    that would be nice, but presumes we can only ever grab a single thing at a time (which i think makes sense?)
- [ ] general cleanup and document all public methods, most importantly in the `_view.py` protocols

at that point, I think I'd like to take a finer-tooth comb through it and make direct edits for a PR to your PR 

### don't need here:

-  pygfx backend.  We can do that later, and until then we can simply refuse to show a histogram in the pygfx version.
- Qt (or any other front-end) wrapper.  I think you can just remove the QtHistogramView.  For the example, just use `VispyHistogramView.view()`, and you can add those additional buttons in some example if you want, but i don't think we (yet) need to have that kind of object in the library.
- don't need to hook this up to the viewer yet. let's do that in another pr.  This should just introduce a stand-alone widget.

src/ndv/histogram/_vispy.py
```diff
@@ -0,0 +1,622 @@
...
+        else:
+            x, y = self._to_plot_coords(event.pos)
+            x1, x2 = self.plot.xaxis.axis.domain
+            y1, y2 = self.plot.yaxis.axis.domain
+            if (x1 < x <= x2) and (y1 <= y <= y2):
+                self._canvas.native.setCursor(Qt.CursorShape.SizeAllCursor)
+            if self._vertical:
+                self._domain = self.plot.yaxis.axis.domain
+            else:
+                self._domain = self.plot.xaxis.axis.domain
```
tlambert03, line 602: why are these lines necessary?  I find that with them, when I move the mouse around the canvas, the domain creeps larger, (causing the graph to continually change).  If I comment them out everything appears fine.  but not sure what the goal was.

gselzer:
> * I would expect the slider to just stop and not go any farther (and pin the newlims to the min/max of the other value accordingly)... but it kinda looks like you're auto-switching the grabbed handle to the other limit?

Yup, this is what I'm doing. I'm not sure what you mean by pinning the newlims to the min/max of the other value, and you could definitely convince me to change the behavior if you can maybe elaborate a bit more on what you want to do and why. 

The situation I'd like to avoid is struggling to grab the desired handle when the sliders overlap (this is what happens, for example, in the fastplotlib histogram). It's possible that we could solve this, as well as implement the behavior you're interested in, by making explicit handles that are dragged instead of the line, like what mmstudio offers.

> need tests. I'm not worried about testing the visual/on-screen representation for now. but smoke testing all public methods for exceptions should happen.

:+1: (also goes for testing/documentation/cleanup)

> that would be nice, but presumes we can only ever grab a single thing at a time (which i think makes sense?)

The only situation where I could see wanting to grab multiple things at once is if you wanted to "translate" the clims left/right (i.e. grab both). But we could make a separate enum for that, because you'd presumably want to specify that window with a separate object (like fastplotlib does).

> * pygfx backend.  We can do that later, and until then we can simply refuse to show a histogram in the pygfx version.

Sure, I'll throw my current efforts onto a new branch.

> * Qt (or any other front-end) wrapper.  I think you can just remove the QtHistogramView.  For the example, just use `VispyHistogramView.view()`, and you can add those additional buttons in some example if you want, but i don't think we (yet) need to have that kind of object in the library.

That was actually mostly created as a visual test of the widget, and my plan (aligning with your suggestion) was to move it to an example. Good to have an explicit reminder though!

> * don't need to hook this up to the viewer yet. let's do that in another pr.  This should just introduce a stand-alone widget.

:+1:

src/ndv/histogram/_vispy.py
gselzer: Oh, I'm sorry. I was aware of that bug, and had deleted it locally but had forgotten to push. Removed with c5a5635

gselzer:
@tlambert03 I added handles for adjusting the clims in 89161c4. I think that the Markers could use some manipulation, but this would solve both my frustration of being unable to distinguish between left and right clims when they are overlapped, and your frustration of the clims surprisingly "sliding" past each other. I'm not tied to the idea, but let me know what you think!

image

gselzer:
Did a large amount of documentation/testing/cleanup in 0090abc. Still lots to fix here (including the missing dependencies, which I should have just added in this PR (hadn't yet due to the threat of merge conflicts in #51)

tlambert03 (approved):
this is just lovely.

I pushed some minor tweaks, just some comments/docs and repositioning of a couple things.   My only two thoughts are:

1. it might be possible to combine the 4 nodes (histogram and handles) into a single `vispy.scene.Compound` widget, and get rid of some code... but then you'll still need to figure out how to have a different transform for the handles than that histogram.  so I'm 100% fine with the way you've done it here.
2. I think i do prefer the old clim-grabbing behavior (before the circle handles at the top and the bottom).  While it's nice that they are unambiguous when on top of each other, they're too hard to grab... I want to grab the line itself.  I would not mind at all if, when the two lines are on top of each other, you always just get the clim-max line (95% of the time, I suspect that's what someone will be moving... and if it wasn't what they wanted, they can always zoom in to disambiguate)

Otherwise LGTM!

gselzer:
> * it might be possible to combine the 4 nodes (histogram and handles) into a single `vispy.scene.Compound` widget, and get rid of some code... but then you'll still need to figure out how to have a different transform for the handles than that histogram.  so I'm 100% fine with the way you've done it here.

If you're fine with it, let's just leave it the way it is then 😅 

> * I think i do prefer the old clim-grabbing behavior (before the circle handles at the top and the bottom).  While it's nice that they are unambiguous when on top of each other, they're too hard to grab... I want to grab the line itself.  I would not mind at all if, when the two lines are on top of each other, you always just get the clim-max line (95% of the time, I suspect that's what someone will be moving... and if it wasn't what they wanted, they can always zoom in to disambiguate)

Rewrote history to remove that behavior, I tried my best to ensure your formatting changes made it through.

tlambert03:
great, thanks for that reversion.  i'm noticing now that the tolerance of `3` for whether a handle is grabbed or not appears to be chosen in data-space, rather than canvas/monitor-pixel space.  That will make for problems with small ranges, e.g. a float histogram between 0-1.  see below, where line is on 20, and mouse is around 17:

https://github.com/user-attachments/assets/5f570be4-ea6a-4e81-853e-b190a28de455

src/ndv/histogram/views/_vispy.py
```diff
@@ -0,0 +1,733 @@
...
+                return Grabbable.RIGHT_CLIM
+            if bool(abs(left - x) < tolerance):
+                return Grabbable.LEFT_CLIM
+
+        if self._gamma_handle_pos is not None:
+            gx, gy, _, _ = self._handle_transform.map(self._gamma_handle_pos[0])
+            if bool(abs(gx - x) < tolerance and abs(gy - y) < tolerance):
+                return Grabbable.GAMMA
+
+        return Grabbable.NONE
```
tlambert03, line 683:
```suggestion
    def _find_nearby_node(
        self, event: SceneMouseEvent, tolerance: int = 4
    ) -> Grabbable:
        """Describes whether the event is near a clim."""
        click_x, click_y = event.pos

        if self._clims is not None:
            vmin, vmax = self._clims
            clims_in_canvas = self.node_tform.imap(np.array([[vmin, 0], [vmax, 0]]))
            canvas_clim_min, canvas_clim_max = clims_in_canvas[:, 0]
            # Right bound always selected on overlap
            if bool(abs(click_x - canvas_clim_max) < tolerance):
                return Grabbable.RIGHT_CLIM
            if bool(abs(click_x - canvas_clim_min) < tolerance):
                return Grabbable.LEFT_CLIM

        if self._gamma_handle_pos is not None:
            gx, gy, *_ = self.node_tform.imap(self._gamma_handle_pos[0])
            if abs(click_x - gx) < tolerance and abs(click_y - gy) < tolerance:
                return Grabbable.GAMMA

        return Grabbable.NONE
```

this needs to be something more like the above, where we transform the clims and gamma handles into canvas/click space and then check against the tolerance.  the above code works for the clims, and the x component of the gamma handle... but misses the Y component of the gamma handle (because the vertical extent of the gamma handle is 0-1 ... so I suppose `self._handle_transform.imap` or something needs to be involved)

gselzer: @tlambert03 great catch - rewrote the method in 61b5041, let me know what you think

gselzer: Great suggestion, see 5b23fe8

tlambert03 (approved):
couple more things I noticed here while playing with this:

- clim handles don't appear to work in vertical mode, but the gamma handle does.
- though it might seem a little confusing, I think the Y axis should be inverted in "vertical" mode: with negative values at the top, and positive values at the bottom (making it a 90˚ clockwise rotation of the horizontal plot).  alternatively, the y axis could be on the right side, making it a 90˚ counter-clockwise rotation of the horizontal plot.
- in logarithmic mode, the Y axis also needs to be logarithmic, so that the apparent range of values doesn't change (i.e. if the peak value in the regular histogram is ~130, then it should still be 130 in logarithmic mode).  that means we can't *just* apply the logarithm to the counts themselves without modifying the axis as well.  Not sure if vispy supports this easily...

in all cases, I'm happy to just merge this now and have them fixed in a follow up PR.  I've just invited you to the repo with write access, so feel free to merge this if you want to be done with it and move on to smaller incremental fixes :)

gselzer:
> in all cases, I'm happy to just merge this now and have them fixed in a follow up PR. I've just invited you to the repo with write access, so feel free to merge this if you want to be done with it and move on to smaller incremental fixes :)

Nah, let's clean them up now, while we're still thinking about it:

> * clim handles don't appear to work in vertical mode, but the gamma handle does.

Fixed in 71edf3c

> * though it might seem a little confusing, I think the Y axis should be inverted in "vertical" mode: with negative values at the top, and positive values at the bottom (making it a 90˚ clockwise rotation of the horizontal plot).  alternatively, the y axis could be on the right side, making it a 90˚ counter-clockwise rotation of the horizontal plot.

Fixed in d5b4b6a

> * in logarithmic mode, the Y axis also needs to be logarithmic, so that the apparent range of values doesn't change (i.e. if the peak value in the regular histogram is ~130, then it should still be 130 in logarithmic mode).  that means we can't _just_ apply the logarithm to the counts themselves without modifying the axis as well.  Not sure if vispy supports this easily...

I guess I'm unsure what you mean by this - can you elaborate?

tlambert03:
> such that you still see the actual counts in each bin 

yeah this.  that's fine. future goal :)

> EDIT 2: See [here](https://github.com/vispy/vispy/blob/85c4bf944d4f38acc67c215548d3701adb8d9772/vispy/visuals/axis.py#L37) for proof that it isn't supported 😅. EDIT 3: And [more proof](https://github.com/vispy/vispy/blob/85c4bf944d4f38acc67c215548d3701adb8d9772/vispy/visuals/axis.py#L422). Maybe we should try to push a fix upstream?

sure ...future :)

tlambert03:
tests have broken (in case you didn't notice)

gselzer:
> tests have broken (in case you didn't notice)

Ahh, sorry, I thought I had pushed the fix but instead it got caught up in a local-only WIP commit that added those axis labels in my screenshot above. Pushed now.

gselzer:
> Log scales is [on the roadmap](https://github.com/pygfx/pygfx/issues/739) for `pygfx`, but it will be a while before that's ready.

Oh, that's great!

> I'm curious what additional functionality you require!

It's mostly just the things I described [here](https://github.com/pyapp-kit/ndv/pull/52#issuecomment-2455745508):

>

> Looking at fastplotlib, the library has a [HistogramLUTTool](https://github.com/fastplotlib/fastplotlib/blob/f0f80c6a4deabb0941154584c854f2004b0cc316/fastplotlib/tools/_histogram_lut.py#L27), which is great, however:
> 
>     * It doesn't take a precomputed histogram - maybe we could alter the design of the current `StatsModel` to just pass off the data, but I like the idea of precomputing the histogram.
> 
>     * It's a bit buggy - for example, you can rotate the graphic like I do in that script, but the canvas events don't reflect that, so dragging clims behaves unexpectedly 😅
> 
>     * No gamma curve is unfortunate.

The precomputed histogram feature is the one that's really important for us - the others would likely be simpler inclusions.

> I notice you [subclassed](https://github.com/gselzer/ndv/blob/4896fca7ffad91b31dabe1c01858b0f39b8c4555/z.py#L12-L18) the `PanZoomController` to restrict it to a single axis, you can already do this and more with the PanZoomController. See PR: [pygfx/pygfx#778](https://github.com/pygfx/pygfx/pull/778) (also related issue: [pygfx/pygfx#635](https://github.com/pygfx/pygfx/issues/635)), [example1](https://docs.pygfx.org/stable/_gallery/feature_demo/scene_camlink1.html#sphx-glr-gallery-feature-demo-scene-camlink1-py), [example2](https://docs.pygfx.org/stable/_gallery/feature_demo/scene_camlink2.html#sphx-glr-gallery-feature-demo-scene-camlink2-py)

This is also good information, I'll have to take a look!

> Unrelated to this PR, I notice you're also working on ROI tools with `pygfx`, we've got a [RectangleSelector](https://github.com/fastplotlib/fastplotlib/pull/576) but we might take some of what you guys come up with if you have plans for a [polygon and lasso tool](https://github.com/fastplotlib/fastplotlib/issues/142).

Yes, moving things upstream would be great. I know ROI that has implications for the histogram is a line ROI, as in MicroManager we often want to [compute histograms on points along a line](https://micro-manager.org/Version_2.0_Users_Guide#line-profile) - I didn't see that on your issue, but maybe it's worth adding?

PR #54, opened 2024-11-15
feat: ROI model/view, v2

gselzer:
This PR provides a ROIModel class to maintain state about regions of interest, as well as a PRoiView protocol for views to implement when they want to react to changes in a ROIModel. This PR is by no means ready for merge, but exists for visibility and discussion about what is needed here.

For now, I'll edit mvc.py to include a (rectangular) ROI - next steps are to make it movable/resizable

Running points of discussion:

* What attributes are needed for the ROIModel? For a ND rectangular ROI, all we really need is minimum and maximum points in N dimensions, but thinking more broadly about lines, ellipses, freehand ROI selections, we might want more here. One idea I kind of like is an attribute that can provide a field of positions - something like imglib2's IterableRegion.
* Should ROIModel itself be a protocol that e.g. RectangularROIModel, LineROIModel, etc. implement?

### DIFF
--- a/mvc.py
+++ b/mvc.py
@@ -5,6 +5,7 @@ from collections.abc import Sequence
 from ndv import data, run_app
 from ndv.controller import ViewerController
 
+# import os
 # os.environ["NDV_CANVAS_BACKEND"] = "pygfx"
 viewer = ViewerController()
 
@@ -22,6 +23,6 @@ viewer.data = img_data
 viewer.show()
 viewer.model.current_index.update({0: 32, 1: 1})
 viewer.roi.events.bounding_box.connect(foo)
-viewer.roi.bounding_box = ([10, 10], [100, 100])
+viewer.roi.bounding_box = ((10, 10), (100, 100))
 
 run_app()
--- a/mvc.py
+++ b/mvc.py
@@ -1,12 +1,12 @@
 """Example usage of new mvc pattern."""
 
-import os
 from collections.abc import Sequence
 
 from ndv import data, run_app
 from ndv.controller import ViewerController
 
-os.environ["NDV_CANVAS_BACKEND"] = "pygfx"
+# import os
+# os.environ["NDV_CANVAS_BACKEND"] = "pygfx"
 viewer = ViewerController()
 
 
--- a/mvc.py
+++ b/mvc.py
@@ -1,12 +1,11 @@
 """Example usage of new mvc pattern."""
 
-import os
 from collections.abc import Sequence
 
 from ndv import data, run_app
 from ndv.controller import ViewerController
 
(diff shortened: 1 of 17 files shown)


gselzer:
@tlambert03 one fun bug that just popped up for me is some strange rendering behavior if you run `mvc.py`, but then use the roi button to create a new ROI - parts of the rectangle border do not render (and the specific ones that do changed based on the window zoom), and the handles are occluded by the rectangle...kinda?

I'd be interested to see if you can reproduce 😅 

python_oE1INx9Fq4

tlambert03:
> I'd be interested to see if you can reproduce 😅

nope... wasn't able to see that

tlambert03:
would love to see this merged into v2 now.  The recent mouseable stuff (#64) will probably be relevant here.  there's likely some patterns to emulate now

mvc.py
```diff
@@ -1,15 +1,28 @@
...
+
 img_data = data.cells3d()
-img_data = np.tile(img_data, (1, 1, 3, 4))
+# img_data = np.tile(img_data, (1, 1, 3, 4))
 viewer.data = img_data
 viewer.show()
 viewer.model.current_index.update({0: 32, 1: 1})
+viewer.roi.events.bounding_box.connect(foo)
+viewer.roi.bounding_box = ((10, 10), (100, 100))
+
```
gselzer, line 27: Should clean this up before merge.

src/ndv/controller/_controller.py
```diff
@@ -140,6 +160,25 @@ def _set_model_connected(
         ]:
             getattr(obj, _connect)(callback)
 
+    # FIXME
+    def _set_roi_connected(
+        self, model: RectangularROIModel, connect: bool = True
+    ) -> None:
+        """Connect or disconnect the model to/from the viewer.
+
```
gselzer, line 168: Probably warrants fixing when we actually add support for multiple ROIs

gselzer:
@tlambert03 I think this is ready for your eyes and testing.

I refactored the ROI design to be focused on (a single) rectangular ROI, as that's where my current needs are. I tried to make the design easily refactored to account for multiple ROIs. Ideally we could also generalize to multiple ROI types, but that might be a bit more work.

Additionally, this PR has an `ArrayViewerModel`, which is designed to capture viewer state. We had discussed making this a separate PR, however I really have no needs for it beyond keeping track of ROI behavior, so it would have made for a pretty worthless PR if I had split it out 😅. One additional ROI characteristic this might soon keep track of is the maximum amount of ROIs allowed on the canvas (likely another enum, with "one", "many", or "none" as options).

My current TODO list:
* [ ] Write tests.
* [ ] Ensure remaining occurrences of "bounding box" are correct (I did some renaming...)

PR #55, opened 2024-11-22
feat: port histogram model

gselzer:
This PR ports the work done in #52 to the `v2-mvc` branch, performing significant restructuring of the business logic. It does not (or shouldn't 😅) integrate with the existing model/view (yet), although the goal is to make it trivial to do so later via modular components.

Notably, this PR also currently includes support for Jupyter (no pygfx yet though 😢)

firefox_NeXfRegyJY


src/ndv/views/_jupyter/jupyter_view.py
```diff
@@ -166,3 +167,143 @@ def show(self) -> None:
...
+        self._backend.setGamma(gamma)
+        self._backend.refresh()
+
+    def setClims(self, clims: tuple[float, float] | None) -> None:
+        self._backend.setClims(clims)
+        self._backend.refresh()
+
+    def setAutoScale(self, autoscale: bool | tuple[float, float]) -> None:
+        self._backend.setAutoScale(autoscale)
+        self._backend.refresh()
```
tlambert03, line 267: the fact that both jupyter_view and qt_view have all these methods, and all both of them are doing is controlling the same backend object is a good indication that this should be controller logic, not front-end view logic.  I suspect you were probably just mirroring how I have `get_canvas_class` in my `QViewerView`, but that was a temporary pattern (it has a note on it about moving it to the controller) and has been removed in my current branch.

kinda makes me feel like, in the interest of not wasting your time, i finalized some of those (very experimental) pattern in my branch before you start to extend them for other mvcs? 😬

src/ndv/views/_jupyter/jupyter_view.py
```diff
@@ -166,3 +167,143 @@ def show(self) -> None:
...
+
+    def set(self, type: CursorType) -> None:
+        if type is CursorType.DEFAULT:
+            self._native.cursor = "default"
+        elif type is CursorType.V_ARROW:
+            self._native.cursor = "ns-resize"
+        elif type is CursorType.H_ARROW:
+            self._native.cursor = "ew-resize"
+        elif type is CursorType.ALL_ARROW:
+            self._native.cursor = "move"
```
tlambert03, line 309: i'm not convinced for the need for `get_cursor_class` and this yet.  I do like that you've added a new front-end agnostic `CursorType` enum concept.  but it might be better to just have the front_ends objects to have a `set_cursor(self, cursor_type: CursorType)` method, rather than this object that maintains a pointer to some native object so that it can mutate it.  need to keep looking...

src/ndv/views/_jupyter/jupyter_view.py
tlambert03: > this should be controller logic, not front-end view logic.

as an exercise, you should try to use the same controller for both the mvc_histrogram.py and notebook examples.  thats' where most of this should go I think...

tlambert03 (commented):
hey @gselzer, good to see some of this coming together!

I've listed a few thoughts.  I don't all that convicted one way or the other about anything, but there are just some vague "smells" at this point that I need to keep thinking about.  It all feels very abstracted and verbose.   I know it very much models the patterns I'm playing with in the mvc branch, but I'm still not quite sure it needs that much.  

One thing that's worth noting: if the histogram and all of its interactive elements like clims handles and stuff are purely vispy/pygfx objects, then we don't actually need Qt and Jupyter wrappers at all.  That is already provided for us by vispy & jupyter_rfb directly.  We only need these qt/jupyter views when we're adding additional sliders and buttons that are also controlling the model

src/ndv/models/_stats_model.py
```diff
@@ -0,0 +1,138 @@
...
+        cls, _core_schema: core_schema.CoreSchema, handler: GetJsonSchemaHandler
+    ) -> JsonSchemaValue:
+        # Use the same schema that would be used for arrays
+        return handler(core_schema.list_schema(core_schema.any_schema()))
+
+
+NumpyNdarray = Annotated[np.ndarray, _NumpyNdarrayPydanticAnnotation]
+
+
+class StatsModel(NDVModel):
```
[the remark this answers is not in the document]
gselzer: > So, it's possible we could reduce the complexity here a bit by not treating this as a first class object that we pass in to `__init__` functions and stuff, but rather just as an implementation detail.

Well, I do think that it has the smallest case for being a true `Model` object, and I do agree that it maybe should not be passed to `__init__` functions, but I'm also hesitant to remove the `Model` status, because having the signals is useful and I'm not sure how the signals would work without an object to emit them. I'll continue thinking about this as I evolve the design.

src/ndv/views/_jupyter/jupyter_view.py
gselzer: Is the `front_ends` object something new from #56?

gselzer:
Thanks for taking a look @tlambert03!

> One thing that's worth noting: if the histogram and all of its interactive elements like clims handles and stuff are purely vispy/pygfx objects, then we don't actually need Qt and Jupyter wrappers at all. That is already provided for us by vispy & jupyter_rfb directly. We only need these qt/jupyter views when we're adding additional sliders and buttons that are also controlling the model

As I mentioned above, the only reason those versions exist right now is to add the `logarithmic range`/`vertical domain` buttons that were added in the v1 example. I do think that we will want non-canvas controls for our histograms, so I'm hesitant to remove them, but at the same time I get where you're coming from. Maybe I'll have more ideas once I dive into #56 and rebase this PR on top...

src/ndv/models/_stats_model.py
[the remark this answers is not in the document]
tlambert03: Yeah signals are definitely important, but there are many ways to give something signals without bringing pydantic and full model status into the mix.  Have a look at basic example in psygnal readme

src/ndv/views/_jupyter/jupyter_view.py
tlambert03: i started using the term "front-end" for the GUI's (qt, jupyter)

src/ndv/models/_stats_model.py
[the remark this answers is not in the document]
gselzer: Upon talking more with @marktsuchida about this, I restructured `StatsModel` down to a dataclass in c9c9be2. It's possible we don't even need *that*, and that the `PStatsView` protocol is also unnecessary. I'll think about it more next week - opinions welcome in the meantime!

src/ndv/views/_jupyter/jupyter_view.py
gselzer: In 05a9aee, I created a `HistogramController` class. It's mostly just the `Controller` class I had in the example `mvc_histogram.py`, with some additional features. `JupyterHistogramView` and `QtHistogramView` now exist as widget "frontends", aligning with the `PView` protocol. `VispyHistogramCanvas` is now implementing the `PHistogramCanvas` protocol as a histogram "backend" - ideally a pygfx version will also exist.

src/ndv/views/_jupyter/jupyter_view.py
gselzer: This behavior has now been migrated into the `PHistogramView` "frontend" widgets. No more need for the `PCursor` implementations (one of which being the `JupyterCursor` discussed here)

gselzer:
Some remaining discussions, all of which I think are non-blockers that we could file issues for (reflecting the rapid iteration goal of the `v2-mvc` branch:
* [ ] I find sporadic [segmentation faults](https://github.com/pyapp-kit/ndv/actions/runs/12206023351/job/34054624108#step:15:50) in CI, arising from `tests/test_nd_viewer.py::test_ndviewer`. I'd be surprised if this is related to my changes...
* [ ] We should add some (public) API (at some points) for the `PHistogramView` methods for setting domain, range, domain axis, etc. While the histogram widget added in #52 had a button "Vertical" for setting the domain axis. Options include:
  * Access to the `PHistogramCanvas` via a public field or property.
  * Wrapper methods on the `HistogramController`.
  Either way, we can add that later.
* [ ] We may want some API to disable pan/zoom - maybe `PHistogramCanvas.set_interactive(enabled: bool)`?
* [ ] We should decide on what it means to "reset zoom". 
  * Should the zoom center around the data, or lut controls as well? Right now it's both, which I think I like.
  * There should probably be a margin on either side of the domain range, as right now if a clim is on the edge, it essentially becomes invisible:
image

* [ ] There is a bug around setting logarithmic range where the domain also gets reset to the data extents - this would ideally be fixed before merge.
* [ ] In logarithmic range, ideally the axis would scale logarithmically, instead of us computing the base-10 log of each bin height. But vispy doesn't support it (yet)

The only other TODO on my plate right now is increasing test coverage on the `PHistogramView` implementations and the `HistogramController`. I think that the API design is looking pretty good right now though!

tlambert03:
heads up @gselzer, I pushed my conflict resolution in 102402c to this branch.  the intention wasn't to change anything here, but have a look just in case (and don't forget to pull before making changes)

tlambert03:
thank you again, your work here is included in a joint commit in #65

PR #56, opened 2024-11-23
updates to v2-mvc branch

tlambert03:
just tracking some of the changes I'm making on the v2-mvc branch.   this is in continuing with #51 , and for comparison with the parts of the codebase that @gselzer  is touching in #54


tlambert03:
@gselzer, this is in a nice state right now and I will probably merge within the day.

it touches a lot of things, but mostly just refactors.  It adds back text info above and below the canvas, and adds back the hover value pattern.  Removes the `get_canvas()` from the Qt/Jupyter side of things, and has the controller control the canvas directly.  Sets up a testing pattern where one can use `NDV_GUI_FRONTEND='qt' pytest` or `NDV_GUI_FRONTEND='jupyter' pytest` to specifically test one gui frontend... (but doesn't yet use it on CI, since I need to move away from using the reusable workflow that I use).

I want to now work on the data request loop and channel splitting, and will start in a new branch

PR #62, opened 2024-12-08
feat: add Wx front-end

tlambert03:
@gnodar01, we discussed this a while back, so thought you might like to know that some progress has been made on the wx-support front.  Some recent changes here (clearer model view separation) have made it easier for us to support arbitrary front-ends (qt, jupyter, and here wx) with relatively minimal additional work:

https://github.com/user-attachments/assets/b3a991ce-6715-4ec5-ae98-9e3ff9d628dd

clearly I'm not particularly skilled at styling things in wx :joy: ... but the bones are there!


gnodar01:
Super cool @tlambert03! I'll take a deeper look this week to see how much friction is involved in integrating into CellProfiler.

tlambert03:
cool.  over the next week or so, I'll be doing some more changes.  and will try to make it clearer exactly *how* you are supposed to use the thing (so you're not searching too hard).  And then, it would be great to get your wish-list for things that aren't yet implemented that would be particularly helpful or necessary for your main use cases

tlambert03:
note to self: the `_WxArrayViewer` isn't properly inserting the canvas widget for the pygfx backend.  the fix is to use `canvas_widget._subwidget` ... but I want to look into a more robust fix

tlambert03:
ok, this is going in! :) many styles and improvements can come.  it's nice just to have the framework for wx in there for now.  @gnodar01, I'll ping you once there's a pre-release on pypi for you to try. so you can just `pip install 'ndv[wx]' --pre` and play with it

PR #66, opened 2024-12-11
refactor: v2 - use base classes rather than protocols

tlambert03:
This is a rather invasive PR that moves from uses Protocols to Base classes for all of the canvas/gui stuff

I have become increasingly frustrated with how it *feels* to use protocols.  They are hard to reason about and hard to refactor using standard IDE tools (since the connection between the protocol and the concrete implementation is much more of an abstract "contract" rather than any sort of direct connection).   As the number of protocols has grown, they've also become partially overlapping and use different names for similar concepts  (@gselzer  has mentioned this elsewhere too).


src/ndv/views/bases/graphics/_canvas_elements.py
```diff
@@ -0,0 +1,94 @@
...
+
+class ImageHandle(CanvasElement):
+    @abstractmethod
+    def data(self) -> np.ndarray: ...
+    @abstractmethod
+    def set_data(self, data: np.ndarray) -> None: ...
+    @abstractmethod
+    def clim(self) -> Any: ...
+    @abstractmethod
+    def set_clim(self, clims: tuple[float, float]) -> None: ...
```
gselzer, line 71: `set_clims` for uniformity with e.g. `LutView.set_clims`?

src/ndv/views/bases/_lut_view.py
```diff
@@ -0,0 +1,78 @@
...
+    def set_gamma(self, gamma: float) -> None:
+        """Set the gamma value of the LUT."""
+        return None
+
+    # These methods apply a value to the view without re-emitting the signal.
+
+    @final
+    def update_auto_scale(self, auto: bool) -> None:
+        with self.autoscaleChanged.blocked():
+            self.set_auto_scale(auto)
```
gselzer, line 58: I got confused when I first saw this in the impls above because I figured `update` and `set` meant the same thing. Can we add some docstrings here to convey the same idea that the comment right above this method is showing? Might also help to consider a prefix like `sync` or `propagate` over `update`, which implies that the change is not "new"

src/ndv/views/bases/graphics/_canvas_elements.py
```diff
@@ -0,0 +1,94 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+    import cmap as _cmap
+    import numpy as np
+
+    from ndv._types import CursorType
+
+
+class CanvasElement(Mouseable):
```
gselzer, line 17: Might make sense for histogram interactive elements to implement this. Could help us theoretically move mouse interactions out of the canvas backends. Would prefer filing an issue for later over doing it here, though

src/ndv/views/bases/graphics/_canvas.py
```diff
@@ -0,0 +1,92 @@
...
+        color: cmap.Color | None = None,
+        border_color: cmap.Color | None = None,
+    ) -> RoiHandle: ...
+
+
+class HistogramCanvas(GraphicsCanvas, LutView):
+    """A histogram-based view for LookUp Table (LUT) adjustment."""
+
+    # TODO: Remove?
+    def refresh(self) -> None: ...
```
gselzer, line 68: Can be removed, already comes from `GraphicsCanvas`

gselzer (approved):
This is super awesome! I have no major complaints - I think everything is pretty minor!

src/ndv/views/bases/graphics/_canvas_elements.py
tlambert03: Yaaas

src/ndv/_types.py
```diff
@@ -71,3 +79,31 @@ class MouseReleaseEvent(NamedTuple):
...
+    BDIAG_ARROW = "bdiag_arrow"
+    FDIAG_ARROW = "fdiag_arrow"
+
+    def apply_to(self, widget: Viewable) -> None:
+        """Applies the cursor type to the given widget."""
+        native = widget.frontend_widget()
+        if hasattr(native, "setCursor"):
+            cast("QWidget", native).setCursor(self.to_qt())
+
+    def to_qt(self) -> Qt.CursorShape:
```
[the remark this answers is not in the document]
tlambert03: `apply_to` uses `to_qt`, and `_vispy._histogram` uses `apply_to`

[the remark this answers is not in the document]
gselzer: Aha, oopsie 😅

src/ndv/views/bases/_lut_view.py
tlambert03: went with the painfully verbose `set_x_without_signal` for now :)

tlambert03:
i think all the review issues are hit here now.  all good for merge @gselzer?

src/ndv/views/bases/graphics/_canvas_elements.py
tlambert03: #67

gselzer:
Yeah, merge it!

PR #73, opened 2024-12-15
ci: test lots of backends on ci

tlambert03:
trying to set up testing of all the things we want to support


.github/workflows/ci.yml
```diff
@@ -23,28 +23,76 @@ jobs:
...
+            canvas: vispy
           - os: ubuntu-latest
-            python-version: "3.11" # unknown CI segfault...
+            python-version: "3.13"
+            gui: jup
+            canvas: pygfx
+          - os: ubuntu-latest
+            python-version: "3.9"
+            gui: pyqt
+            canvas: vispy
```
gselzer, line 51: If the goal is to test some particular use case, it would be nice to document why that use case is worth being included (which could make it easier for us to remember why it was here when it eventually gets removed)

gselzer: I think it would be, but more importantly, is there a reason we don't just set the matrix to run on 3.9 and 3.13 instead of 3.10 and 3.12? In my experience version breakages are usually cumulative (i.e. if it breaks on 3.12 then it also breaks on 3.13)

tlambert03: we could do 3.9, but 3.13 is too soon.  Most of the upstream libraries aren't fully supporting 3.13 yet

gselzer: > Most of the upstream libraries aren't fully supporting 3.13 yet. and that's not really _our_ issue. So what we're doing here is ensuring that our code runs with 3.13 ... and waiting for upstream libraries to catch up before using that as the standard test

Aha, now *that's* what I'd find useful in a message, because that's not something I've been bitten enough by to be thinking about.

gselzer: Yeah, that sounds great. Thank you for dealing with my thick-headedness!

tlambert03: yep ... pyside breaks as usual :joy: https://github.com/pyapp-kit/ndv/actions/runs/12356058279/job/34481082990?pr=73

gselzer: > yep ... pyside breaks as usual 😂 https://github.com/pyapp-kit/ndv/actions/runs/12356058279/job/34481082990?pr=73

Maybe just update the `exclude`?

gselzer: Gotcha, yeah that all makes sense to me. I guess this is exactly the kind of substance that I was hoping would be noted via comments, because I'd hate to forget about all of this trial & error you've done.

PR #75, opened 2024-12-19
feat: add CanvasProvider and GuiProvider classes

### LINKED ISSUE #68: Bug: Explicitly specifying front-end can fail to start app
```sh
$ NDV_GUI_FRONTEND=qt NDV_CANVAS_BACKEND=pygfx python mvc.py
QWidget: Must construct a QApplication before a QWidget
```

this is because this part of the code skips the "try start app" if explicitly provided:

https://github.com/pyapp-kit/ndv/blob/f5ce7e625d44284e110b9bab19ff87980726d45a/src/ndv/views/_app.py#L137-L144

I think the best fix here is to use a "Frontend"  class design, like what @gselzer was proposing elsewhere (can't find the link at the moment).  The only slight variant from what was proposed there is that I think these frontend classes should all be defined in `_app.py`, not in their respective backend folders.  These classes should be able to check the presence/availability of the backends and then do the importing 

```python
class GuiFrontend:
    id: str

    def is_running() -> bool: ...
    def try_start() -> bool: ...
    def exec() -> None: ...
```

similar for canvas backends:

```python
class CanvasBackend:
    id: str

    def is_imported() -> bool: ...
    def is_importable() -> bool: ...
    def array_canvas_class() -> type[ArrayCanvas]: ...
    def histogram_canvas_class() -> type[HistogramCanvas]: ...

```

tlambert03:
closes #68

sorry @gselzer  ... this was blocking some other stuff so I wanted to get it done.    Basic idea here is that, in order to add support for a new gui or canvas provider, you subclass `GuiProvider` or `CanvasProvider`, and then add it to `GUI_PROVIDERS` or `CANVAS_PROVIDERS`.  Then the class methods are used rather than the chain of conditionals


gselzer:
This looks great, no sorry needed!

At some point I think it'd be great if we moved the mouse behavior into these objects to encapsulate all of the frontend logic within a single object. I'm not sure that's convenient/possible with the current design, though, so maybe just make it an issue for later?

tlambert03:
oh right, good point I forgot i [wanted to do that here](https://github.com/pyapp-kit/ndv/issues/68#issuecomment-2538993221) too.  added in 
837ef4c

PR #83, opened 2025-01-09
feat: return 3d support to v2

tlambert03:
reopening #71 against main


tlambert03:
ok @gselzer, this now does the thing I wanted it to (as far as the end user is concerned).  I didn't do too much stuff to have to make it work (though I did pass `_ArrayDataDisplayModel` to all the views... which can be changed later).  Let me know if you can live with it for now

src/ndv/models/_data_display_model.py
```diff
@@ -40,6 +40,8 @@ def __post_init__(self) -> None:
         self.dtype = self.data.dtype
 
 
+# NOTE: nobody particularly likes this class.  It does important stuff, but we're
```
gselzer, line 43: Guilty as charged 🙈

gselzer:
One small thing I noticed while testing - the histogram doesn't update when you go to 3D. If you bring [this `if` block](https://github.com/pyapp-kit/ndv/pull/83/files#diff-38c1067806f281b1362184c2f9210db135547788fab4def7b269cb510802041aR368) outside of the `else` I think things should work as intended.

> Let me know if you can live with it for now

I can definitely live with it in the short term if it gets things working, since everything you changed is private API, no?

PR #85, opened 2025-01-10
Docs deploy

tlambert03:
setup versioned docs deployment


tlambert03:
docs are now live at https://pyapp-kit.github.io/ndv/dev/  (until we actually have a new tagged commit ... at which point you can see it at https://pyapp-kit.github.io/ndv/

Issue #70, opened 2024-12-14
consider passing the model directly to gui view objects


tlambert03:
currently, GUI views don't touch the model... they just notify the controller of a change, and then the controller requests the changed value from the view before passing it to the model:

```python
class SomeView:
    currentIndexChanged = Signal()
    def __init__(self):
qwdg.dims_sliders.currentIndexChanged.connect(self.currentIndexChanged.emit)
    def current_index(self): ...

class Controller:
    def __init__(self):
        self.model = ...
        self.view = ...
        self.view.currentIndexChanged.connect(self._on_view_current_index_changed)

    def _on_view_current_index_changed(self) -> None:
        self.model.current_index.update(self.view.current_index())
```

That works fine, but it requires a lot of signal machinery to be on the View objects, and adds some indirection.   We could alternatively just pass the model to the frontend widget:

```python
class SomeView:
    def __init__(self, model):
        self.model = model

    def current_index(self): ...

class Controller:
    def __init__(self):
        self.model = ...
        self.view = ...
        self.view.currentIndexChanged.connect(self._on_view_current_index_changed)

    def _on_view_current_index_changed(self) -> None:
        self.model.current_index.update(self.view.current_index())
```

tlambert03:
on a quick stab at implementing it, I ran into one other issue that would likely need addressing:  the view has the ability to reset the range of the canvas.  Ideally this would actually correspond to a change on the model, but we don't yet model the camera in the display model.    That's a big ball of wax that was a real sticking point in microvis ... but we'll need to tackle it someday

Issue #47, opened 2024-10-08
Add FPS Display


gselzer:
### Description

Especially for a live data feed (think Live view from Micromanager), it might be nice to display the FPS somewhere on the window.

PR #87, opened 2025-01-12
refactor: Pass LUTModel to LUTView

gselzer:
This PR is designed to simplify the current MVC architecture of the lookup tables. This work was originally part of a larger PR that does similar things to all of the model-view pairings, but there's no reason it cannot stand on its own and be scrutinized earlier. In other words, it's nice to discuss whether this is a good idea before applying this design to the rest of the codebase.

As noted in #70, the resulting design is simpler, removing signals from the LUTView objects. They instead maintain access to a `LUTModel` object, which is updated directly. Much of the boilerplate logic in the `ChannelController` thus goes away, and it simply becomes an organizer for one `LUTModel` and many `LUTView`s (which in practice is a LUT widget, an image handle, and sometimes a histogram)

One additional benefit here is that testing `LUTModel` and `LUTView` should become much easier without the controller being needed. I'd like to add some of these tests as a part of this PR, once we decide we like the changes.

TODO:
* [x] Reconcile `ImageHandle` and `LUTView`. It seems to me like `ImageHandle` should be a `LUTView`, however there are a few methods brought in from the `Viewable` ABC that are strange here. There's a bit of naming skew between them though, which would be nice to resolve.
* [x] There are many `if self.model` checks, which I find annoying. They exist because the `LUTView` does not require the model at construction time - this would require us pass a `LUTModel` to functions on e.g. an `ArrayView`, which I found strange. But maybe that's less annoying? 
* [ ] Clean
* [x] Test


tlambert03:
i took the liberty of fixing merge conflicts (which were caused by my docs stuff in #84)

src/ndv/views/bases/_lut_view.py
```diff
@@ -1,20 +1,19 @@
...
 
-    visibilityChanged = Signal(bool)
-    autoscaleChanged = Signal(bool)
-    cmapChanged = Signal(cmap.Colormap)
-    climsChanged = Signal(tuple)
-    gammaChanged = Signal(float)
+    _model: LUTModel | None = None
+
+    def __init__(self, model: LUTModel | None = None) -> None:
+        self.model = model
```
tlambert03, line 16: I believe `__init__` is actually unused and can be removed (i.e. nowhere do we actually pass in the model during `__init__`)

tlambert03 (commented):
this is great @gselzer.  I _really_ like the direction this is heading!

There's one important thing that's been lost though (and it looks like I hadn't added a test for this yet)  and that is the synchronization between auto-scale and clims:
1. manually setting the clims needs to turn off autoscaling
2. manually toggling autoscale (when it was previously off) needs to update the clims slider

https://github.com/user-attachments/assets/f89760a0-73b6-49f5-a5bc-7608b56b6a0e

this is a somewhat tricky thing to actually do, because it can cause circular signals (e.g. setting `autoscale=True` on the model will set clims to the autoscaled version ... but that should *not* result in `model.autoscale=False`.  By contrast, if the *user* sets clims (as opposed to the action of autoscale), then we do need to set `model.autoscale = False`).  For example, on `main`:

https://github.com/user-attachments/assets/e0a78c29-f36c-4db8-95bf-77ba06edf495

one thing I fear may happen with this PR is that that complicated behavior now needs to be re-implemented on each frontend view?   whereas those `set_without_signal` methods were previously handling that for all views (but I'm not sure yet)

tlambert03:
> it looks like I hadn't added a test for this yet

if it would help, I can try to add this test to main?

gselzer:
> There's one important thing that's been lost though (and it looks like I hadn't added a test for this yet) and that is the synchronization between auto-scale and clims:
> 
>     1. manually setting the clims needs to turn off autoscaling
> 
>     2. manually toggling autoscale (when it was previously off) needs to update the clims slider

Yeah, good catch, let me try to work this in tomorrow.

> one thing I fear may happen with this PR is that that complicated behavior now needs to be re-implemented on each frontend view? whereas those `set_without_signal` methods were previously handling that for all views (but I'm not sure yet)

Well I hope a lot of the logic can go within the `LUTView` class itself. Some of the autoscale behavior could be tricky and may need to remain within the controller (or maybe it could live within the `ImageHandle`?), but we should be aware of whether this design is actually worth it as we work out the remaining kinks.

> > it looks like I hadn't added a test for this yet
> 
> if it would help, I can try to add this test to main?

If you want to, and it's not a lot of effort, then sure!

gselzer:
> There's one important thing that's been lost though (and it looks like I hadn't added a test for this yet) and that is the synchronization between auto-scale and clims:
> 
>     1. manually setting the clims needs to turn off autoscaling
> 
>     2. manually toggling autoscale (when it was previously off) needs to update the clims slider

This has been fixed with 9abec15.

> one thing I fear may happen with this PR is that that complicated behavior now needs to be re-implemented on each frontend view? whereas those `set_without_signal` methods were previously handling that for all views (but I'm not sure yet)

Yeah, I mean there's a little bit - let me know if you think the changes are not worth it. I think that the code added to jupyter is the worst - if you know of a better way to handle that code, let me know!

It's worth noting that for wx, it seems like the default `SetValue` behavior does not fire off events (see [`ToggleButton`](https://docs.wxpython.org/wx.ToggleButton.html#wx.ToggleButton.SetValue), [`CheckBox`](https://docs.wxpython.org/wx.ToggleButton.html#wx.ToggleButton.SetValue)), so I removed the event code from the `RangeSlider`. Notably,  I had trouble testing it, because I can't see the range slider:

image

src/ndv/views/bases/_lut_view.py
gselzer: Yup, removed it. Thanks!

src/ndv/views/bases/graphics/_canvas_elements.py
```diff
@@ -74,9 +76,27 @@ def gamma(self) -> float: ...
...
+    def frontend_widget(self) -> Any:
+        return None
+
+    def set_channel_name(self, name: str) -> None:
+        pass
+
+    def set_auto_scale(self, checked: bool) -> None:
+        if checked and self.model:
+            d = self.data()
+            self.model.clims = (d.min(), d.max())
```
gselzer, line 96: Note to self - this behavior should probably go back into the controller for this PR. Otherwise, we're going to have a strange situation when multiple handles are active on a single channel.

gselzer: Fixed.

gselzer:
@tlambert03 I think this is ready for your eyes again. Points of discussion:

* Again, we need to decide whether we like this general paradigm, where:
  * Each view listens to its model directly. The `LutView` methods `set_auto_scale`, `set_clims`, etc. should take care to not alter the model in their execution.
  * Updates to any view directly alter the model, meaning the updates of one view then result in model signals listened to by all models
  * The controller ideally just connects/syncs the model(s?) and views.
    * Right now, the `ChannelController` does more than this, but that's okay for this PR. Particularly, the remaining behavior contains functionality linking the image data to the model, but I'd prefer it to be moved out later if we like this pattern.
* The additional tests are useful and found at least one bug, let me know what you think. The pyside tests are still segfaulting for me at times, if you have any quick intuition there I'd be grateful, and it looks like a wxPython test is failing too. Still working on that part, but it shouldn't postpone your review.
* The `JupyterLutView` changes using the `_updating` field are ugly, which is unfortunate. It's also unfortunate that all three views are turning off autoscaling when their clims object changes. Right now I think it's manageable, but maybe if we have to write more logic in the views we should move this functionality to `LutView`.

tlambert03:
just took another look here.  As mentioned in conversations on zulip, I think the current deal breaker for me here is what this change, as written, implies for what it means to write views.  It moves us away from a pattern where all one has to do in order to implement a frontend is simply add a few setter methods for each property, and into a place where each view has to carefully manage their own events, in their own pattern, in order to update the model in the correct way.  To me, that opens up both the possibility that different views behave quite differently from each other, and it also puts a bigger burden on what it means to add additional fields to the model, and to add new possible views.  It *does* make the controller now look very nicely slim, but the slimness of that 1 controller comes at the expense of the fatness of all the views.

I definitely agree we haven't yet landed on particularly nice pattern in main either, but i also don't think i'd be happier going in this direction.

as a side-note: it's conceivable that #94 might have removed the need for all of the blockers that are currently in place (and which you are trying hard to maintain in this PR).  The reason would be that it removed the one thing that caused us to need to know *who* was updating the model (i.e. whether it was the view or the user).  With that gone, we may well never have to worry about vicious loops.

gselzer:
> Into a place where each view has to carefully manage their own events, in their own pattern, in order to update the model in the correct way. 

Yeah, that is also the largest drawback for me. It is nice to be able to abstract the different event intricacies behind psygnal.

> To me, that opens up both the possibility that different views behave quite differently from each other and it also puts a bigger burden on what it means to add additional fields to the model, and to add new possible views. 

I don't think I understand what you're worried about here, can you give an example?

The only additional burden here is that the LutViews must avoid side-effects, as we cannot rely on psygnal to avoid them. It is a bit more code, but it is also a cleaner contract and simpler control flow. Adding additional fields to the model should be equally difficult either way (although we could make that easier by adding no-op implementations to the view ABC).

> I definitely agree we haven't yet landed on particularly nice pattern in main either, but i also don't think i'd be happier going in this direction.

I do think that this design simplifies control flow, simplifying debugging and maintenance, but if you feel strongly here then I'm content to move on from this experiment!

src/ndv/views/_jupyter/_array_view.py
```diff
@@ -70,20 +71,30 @@ def __init__(self) -> None:
         self._cmap.observe(self._on_cmap_changed, names="value")
         self._clims.observe(self._on_clims_changed, names="value")
         self._auto_clim.observe(self._on_autoscale_changed, names="value")
+        self._updating: bool = False
```
tlambert03, line 74: looked at this again today... and I do think I want to try to use it.  I think it's mostly these views that stand in my way, so let's just try to clean those up:

in this jupyter file, let's get rid of the `_updating` attribute... and add a new module function:


```python
@contextlib.contextmanager
def notifications_blocked(
    obj: HasTraits, name: str = "value", type: str = "change"
) -> Iterator[None]:
    # traitlets doesn't provide a public API for this
    notifiers: list | None = obj._trait_notifiers.get(name, {}).pop(type, None)
    try:
        yield
    finally:
        if notifiers is not None:
            obj._trait_notifiers[name][type] = notifiers
```

and then *only* add it to these two... since I believe they're the only ones that have the potential for recursion:

```python
    def set_clim_policy(self, policy: ClimPolicy) -> None:
        with notifications_blocked(self._auto_clim):
            self._auto_clim.value = not policy.is_manual

    def set_clims(self, clims: tuple[float, float]) -> None:
        with notifications_blocked(self._clims):
            self._clims.value = clims
```

... and then remove the `not self._updating` from the `_on_x_changed` methods

src/ndv/views/_qt/_array_view.py
```diff
@@ -135,27 +142,30 @@ def __init__(self) -> None:
...
-    def set_auto_scale(self, auto: bool) -> None:
-        self._qwidget.auto_clim.setChecked(auto)
+    def set_clim_policy(self, policy: ClimPolicy) -> None:
+        with signals_blocked(self._qwidget.auto_clim):
+            self._qwidget.auto_clim.setChecked(not policy.is_manual)
 
     def set_colormap(self, cmap: cmap.Colormap) -> None:
-        self._qwidget.cmap.setCurrentColormap(cmap)
+        with signals_blocked(self._qwidget.cmap):
+            self._qwidget.cmap.setCurrentColormap(cmap)
```
tlambert03, line 162:
```suggestion
        self._qwidget.cmap.setCurrentColormap(cmap)
```

(don't think this one needs it)

src/ndv/views/_vispy/_array_canvas.py
```diff
@@ -426,6 +426,7 @@ class VispyArrayCanvas(ArrayCanvas):
 
     def __init__(self) -> None:
         self._canvas = scene.SceneCanvas(size=(600, 600))
+        self._canvas.measure_fps()
```
tlambert03, line 429: remove or make optional

tlambert03 (commented):
ok, I think i'm having a change of heart here :)

I sat down to pick out the parts i liked from the parts i didn't, and honestly really wasn't left with too much besides the front-end views, and even those weren't as much as i thought they were.  (again, apologies for my waffling on this).  I left a suggestion for cleaning up the jupyter pattern, we can likely add one to wx later too)

in addition to the other comments , I think if we added `__eq__` operators to `ClimPolicy`, we would avoid a lot of unnecessary signals   (which only get emitted when the value is not equal to the previous value):

```python
class ClimsManual(ClimPolicy):
    ...
    def __eq__(self, other: object) -> bool:
        return (
            isinstance(other, ClimsManual)
            and self.min == other.min
            and self.max == other.max
        )


class ClimsMinMax(ClimPolicy):
    ...
    def __eq__(self, other: object) -> bool:
        return isinstance(other, ClimsMinMax)


class ClimsPercentile(ClimPolicy):
    ...
    def __eq__(self, other: object) -> bool:
        return (
            isinstance(other, ClimsPercentile)
            and self.min_percentile == other.min_percentile
            and self.max_percentile == other.max_percentile
        )


class ClimsStdDev(ClimPolicy):
    ...
    def __eq__(self, other: object) -> bool:
        return (
            isinstance(other, ClimsStdDev)
            and self.n_stdev == other.n_stdev
            and self.center == other.center
        )

```


would you mind integrating those and then marking this not-draft when ready?  i think it's likely good to go at that point

src/ndv/views/_vispy/_array_canvas.py
gselzer: Dang, my debug code keeps slipping in here. Good catch! I really need to work on #47 and/or some sort of debug mode...

gselzer:
> in addition to the other comments , I think if we added `__eq__` operators to `ClimPolicy`, we would avoid a lot of unnecessary signals (which only get emitted when the value is not equal to the previous value):

Oh, yeah, there are some nice ideas here! Let me take a look!

src/ndv/views/_jupyter/_array_view.py
gselzer: Yup, works great in [aa7ac47](https://github.com/pyapp-kit/ndv/pull/87/commits/aa7ac4743b20aedef65116c9d490615775a5d5fb)

src/ndv/views/_qt/_array_view.py
gselzer: Yeah - it's not the only one. Cleaned up in [2964408](https://github.com/pyapp-kit/ndv/pull/87/commits/2964408fb94941b5b07a385182b42e27a2f5d312)

tlambert03 (approved):
good stuff.  really quite lovely, i don't know what i was smoking before :joy:

only some minor comments

src/ndv/views/_qt/_array_view.py
```diff
@@ -158,27 +165,28 @@ def __init__(self) -> None:
...
     def set_colormap(self, cmap: cmap.Colormap) -> None:
         self._qwidget.cmap.setCurrentColormap(cmap)
 
     def set_clims(self, clims: tuple[float, float]) -> None:
-        if not isinstance(clims, tuple):
-            breakpoint()
-        self._qwidget.clims.setValue(clims)
+        with signals_blocked(self._qwidget.clims):
+            if not isinstance(clims, tuple):
+                breakpoint()
```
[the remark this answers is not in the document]
gselzer: Removed!

src/ndv/views/bases/_lut_view.py
```diff
@@ -1,77 +1,97 @@
...
+from typing import TYPE_CHECKING
 
 from ._view_base import Viewable
 
+if TYPE_CHECKING:
+    import cmap
+
+    from ndv.models._lut_model import ClimPolicy, LUTModel
+
+    pass
```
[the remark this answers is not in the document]
gselzer: Gone!

src/ndv/controllers/_channel_controller.py
```diff
@@ -27,113 +24,34 @@ class ChannelController:
...
+    def __init__(
+        self, key: LutKey, lut_model: LUTModel, views: Sequence[LutView]
+    ) -> None:
         self.key = key
         self.lut_views: list[LutView] = []
-        self.lut_model = model
+        self.lut_model = lut_model
+        self.lut_model.events.clims.connect(self._auto_scale)
         self.handles: list[ImageHandle] = []
+        self.clims: tuple[float, float] = (0, 1)
```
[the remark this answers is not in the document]
gselzer: Yep! Gone now!

src/ndv/views/_wx/range_slider.py
```diff
@@ -86,8 +86,6 @@ def GetValue(self) -> int:
 
     def SetValue(self, value: int) -> None:
         self.value = value
-        # Post event notifying that value changed
-        self.PostEvent()
```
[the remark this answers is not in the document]
gselzer: > interesting... was this an experiment? something you found unnecessary?

I found that in practice this was unnecessary, yeah.

> but is it correct not to post an event?

Good question...I don't think so, but am not sure. Some wx widgets seem to avoid posting events here (e.g. [`wx.ToggleButton`](https://docs.wxpython.org/wx.ToggleButton.html#wx.ToggleButton.SetValue), [`wx.CheckBox`](https://docs.wxpython.org/wx.CheckBox.html?highlight=checkbox#wx.CheckBox.SetValue)), while others are unclear in the docs (e.g. [`wx.Slider`](https://docs.wxpython.org/wx.Slider.html?highlight=slider#wx.Slider.SetValue)). The ambiguity

Fwiw, there are two ways that this slider can be updated:
1. **User click+drag**: This will call `SliderThumb.SetPosition()`, which will post a `wx.EVT_SLIDER` event. That gets handled in `WxLutView._on_clims_changed`, which sets a `ClimsManual` policy on the model. That model update calls `WxLutView.set_clims` via the `ChannelController._auto_scale`, which then calls `ThumbSlider.SetValue`
2. **Model update**: This is basically just a shorter version of (1.), starting from the autoscaling.

Either way, in practice we don't want an event posted. If you want, we can block it, although I'd have to look up how that can be done.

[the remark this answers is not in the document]
tlambert03: yeah, ok, fine by me.  I'm surprised that it ends up for the `.Bind` call that triggers `_on_clims_changed` without this. but i really know very little about wx.  so let's go with it

tlambert03:
the fails are unrelated... gotta figure out how to get us back to robust CI tests.  ... but later.  thanks for this!

PR #92, opened 2025-01-13
feat: adding back async

tlambert03:
This PR makes the DataRequest -> DataResponse cycle asynchronous again.

It adds awareness of a progress spinner to ArrayView.  See here for slow remote zarr data

https://github.com/user-attachments/assets/d76e8562-2f38-4442-acc7-89af42d946b8

Summary:

- each viewer gets a spinner that they can hide/show, with a new ArrayViewer method to control it
- viewers are all async by default (except Jupyter... which remains single threaded).  and `NDV_SYNCHRONOUS` controls it.
- similar to v1, `_update_canvas` now becomes `_request_data` -> `_on_data_response_ready`
- both qt and wx now have a way to ensure that a method is called in the main thread `@ensure_main_thread`


tlambert03:
note to self:

removing this code from this PR, since it seems better not to use Jupyter Viewer with async requests/responses anyway:

```python

    @staticmethod
    def get_executor() -> Executor:
        # return something that looks like ThreadPoolExecutor
        # but works for Jupyter's asyncio loop
        return _AsyncExecutor.instance()


class _AsyncExecutor(Executor):
    """Simple executor that submits tasks to the asyncio event loop.

    Designed to work with Jupyter notebooks, but have the same API
    as concurrent.futures.ThreadPoolExecutor.
    NOTE: async request/response is turned off at the moment in ArrayViewer
    anyway, because Jupyter already has an asynchronous GUI.
    """

    _instance: ClassVar[_AsyncExecutor]

    @classmethod
    def instance(cls) -> _AsyncExecutor:
        if not hasattr(cls, "_instance"):
            cls._instance = cls()
        return cls._instance

    def __init__(self) -> None:
        import asyncio

        self._asyncio = asyncio
        self.loop = asyncio.get_event_loop()

    def submit(
        self, fn: Callable[P, T], /, *args: P.args, **kwargs: P.kwargs
    ) -> asyncio.Future[T]:
        """Submit a callable to the asyncio event loop.

        Wraps the result in a `Future` that provides a `.result()` method.
        """
        # Create a concurrent Future to return

        # Define a task that wraps the function
        async def run_task() -> T:
            logging.info("running task")
            # Check if fn is a coroutine and await it if necessary
            if not iscoroutinefunction(fn):
                # Use asyncio.to_thread for blocking functions
                result = await self._asyncio.to_thread(fn, *args, **kwargs)
            else:
                result = await fn(*args, **kwargs)
            return result  # type: ignore

        # Schedule the async task on the event loop
        future = self._asyncio.run_coroutine_threadsafe(run_task(), self.loop)
        return cast("asyncio.Future[T]", future)
```

Issue #100, opened 2025-01-17
Installation issue (jupyter)


mese79:
Hey!
Regarding the installation command for `jupyter` (from [here](https://pyapp-kit.github.io/ndv/dev/install/)):
`conda install -c conda-forge ndv vispy pyopengl jupyter jupyter_rfb glfw ipywidgets`

I got two errors:
1. `jupyter_rfb does not exist (perhaps a typo or a missing channel);`
    - `jupyter_rbf` should change to `jupyte-rbf` for conda install.

2. `ndv does not exist (perhaps a typo or a missing channel).`
    - Apparently, there is no `ndv` package on the conda-forge (yet?) 

Anyway, I managed to install `ndv` this way:
```bash
conda install -c conda-forge vispy pyopengl jupyter jupyter-rfb glfw ipywidgets
pip install ndv==0.2.0rc0 
```

tlambert03:
thanks for opening the issue @mese79!

indeed, the docs are "early" here (i expected to have it up before anyone was paying attention here).  ndv isn't yet on conda forge.  i can get to that soon.  will leave this open until then

tlambert03:
ndv is now on conda, and @gselzer  update the docs.  so https://pyapp-kit.github.io/ndv/latest/install/ should be correct now

PR #101, opened 2025-01-17
docs: minor installation fixes & cleanup

gselzer:
Couple small fixes addressing part of the problem in #100:
* `jupyter_rfb` -> `jupyter-rfb`
* `glfw` -> `pyglfw` on jupyter

Additionally, extracted the commands for each source for a little DRYness. Arguably wasn't necessary and more could have been done, but the Boy Scout Rule was nagging me


tlambert03:
beautiful ❤️   thanks

PR #105, opened 2025-01-17
Pygfx histogram v2

gselzer:
This is a repeat of #76, ready for eyes again.

There are a few problems I'm aware of, which I'm tempted just solve in separate PRs:
* There's "shimmering" when you pan/zoom the histogram canvas, due to aliasing. PyGFX's antialias solution of `renderer.pixel_ratio` was not sufficient to fix this on my machine.
  * Note that the VisPy histogram also has this issue, but it is only seen when zooming.
* When you change the data, the pan/zoom resets. This is also a problem for the VisPy histogram.

### DIFF
--- a/src/ndv/views/_pygfx/_histogram.py
+++ b/src/ndv/views/_pygfx/_histogram.py
@@ -1,0 +1,481 @@
+from __future__ import annotations
+
+from enum import Enum, auto
+from typing import TYPE_CHECKING, Any
+
+import numpy as np
+import pygfx
+from fastplotlib import Figure
+from fastplotlib.graphics import LineGraphic
+from fastplotlib.graphics._base import Graphic
+from fastplotlib.graphics.selectors import LinearRegionSelector
+from pygfx.controllers import PanZoomController
+
+from ndv.views.bases import HistogramCanvas
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
+    import cmap
+    import numpy.typing as npt
+
+
+class Grabbable(Enum):
+    NONE = auto()
+    LEFT_CLIM = auto()
+    RIGHT_CLIM = auto()
+    GAMMA = auto()
+
+
+class PanZoom1DController(PanZoomController):
+    """A PanZoomController that locks one axis."""
+
+    _zeros = np.zeros(3)
+
+    def _update_pan(self, delta: tuple, *, vecx: Any, vecy: Any) -> None:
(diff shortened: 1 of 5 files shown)


src/ndv/views/_pygfx/_histogram.py
```diff
@@ -0,0 +1,568 @@
...
+        self._size = (600, 600)
+        self._canvas = cls(size=self._size)
+
+        # this filter needs to remain in scope for the lifetime of the canvas
+        # or mouse events will not be intercepted
+        # the returned function can be called to remove the filter, (and it also
+        # closes on the event filter and keeps it in scope).
+        self._disconnect_mouse_events = filter_mouse_events(self._canvas, self)
+
+        self._renderer = pygfx.renderers.WgpuRenderer(self._canvas, show_fps=True)
```
tlambert03, line 102: maybe remove show_fps or guard it behind an env var or something?

tlambert03 (approved):
complete works for me!

nit re fps

src/ndv/views/_pygfx/_histogram.py
gselzer: Oh, good catch - will remove, I'd rather more thoroughly address that across the repo in #47

tlambert03, line 102:
```suggestion
        self._renderer = pygfx.renderers.WgpuRenderer(self._canvas)
```

tlambert03:
added a basic smoke test

tlambert03:
one test still consistently failing with pygfx and pyside on windows: https://github.com/pyapp-kit/ndv/actions/runs/12856369561/job/35842981229?pr=105

not 100% sure it's the histogram test, but it seems like it's happening right about the same time

tlambert03:
was able to reproduce the test segfault on a real windows 3.12, pyside, pygfx about 50+% of the time.  So I do think it's a real bug.  But since we're already only partially supporting pyside, let's go ahead and skip it

tlambert03:
i'm going to merge this, but I think the additional test I added here, though just a simple smoke test, is causing more random test failures.  It could be two things: either something related to the histogram test itself, or simply related to the fact that the new test is actually the first test added that (re-)uses the backend application for a second time.  That does come with a whole host of possible gotchas.  But, since that's unrelated to this PR, and since we were likely to hit it eventually anyway, I'm going to merge this, and we may have to deal with a brief period of test flakiness while I figure that out.

thanks for your work here!

PR #114, opened 2025-01-21
feat: Rectangular ROIs

gselzer:
This PR brings support for (rectangular) ROIs to the new array viewer. This work was updated from the original viewer, and needs to be cleaned up, but does pass current automated tests and works well in my limited manual testing.

TODO:
* [x] Clean - probably lots of cruft from the update
* [x] Test - I remember this being tricky and being why I was motivated to explore a fix to #70, but let's give it another shot.


gselzer:
Still need to clean up the tests, but I think this is ready for your eyes @tlambert03. Points for discussion:
* This PR introduces the `ArrayViewerModel` to maintain gui state - right now, the only field is `InteractionMode`, but this could house other things (maybe camera state?) in future PRs. I don't really like that name, could rename `UIModel`...
* This PR doesn't really account for future needs (i.e. multiple rois, different shapes, etc. as well as a ROI controller to wrangle them) - my thinking was that this could be future PRs. In addition, I was planning on adding in the integrations I want (such as using the ROI to narrow the histogram down to a region) in future PRs.

src/ndv/models/_roi_model.py
```diff
@@ -0,0 +1,39 @@
...
+        Whether to display this roi.
+    bounding_box: tuple[Sequence[float], Sequence[float]]
+        The minimum point and the maximum point contained within the region.
+        Using these two points, an axis-aligned bounding box can be constructed.
+    """
+
+    visible: bool = True
+    bounding_box: tuple[tuple[float, float], tuple[float, float]] = ((0, 0), (0, 0))
+
+    @field_validator("bounding_box")
```
tlambert03, line 28:
```suggestion
    @field_validator("bounding_box", mode='after')
```

mode after is the default, but i like to be explicit because it's an important distinction.  In this case, it means that you can assume that the `bb` argument being passed in is already going to be a valid tuple and you don't need to check again.

src/ndv/models/_roi_model.py
```diff
@@ -0,0 +1,39 @@
...
+class RectangularROIModel(NDVModel):
+    """Representation of an axis-aligned rectangular Region of Interest (ROI).
+
+    Parameters
+    ----------
+    visible : bool
+        Whether to display this roi.
+    bounding_box: tuple[Sequence[float], Sequence[float]]
+        The minimum point and the maximum point contained within the region.
+        Using these two points, an axis-aligned bounding box can be constructed.
```
tlambert03, line 22: still want a bit more clarity here.  type here says `tuple[Sequence[float], Sequence[float]]` but type of the class says it will always be 2 two-tuples.  The "XYness" of the 2-tuple should be documented and minimum and maximum should be made slightly clearer (e.g. top left, bottom right, or other words that make it stupidly clear)

src/ndv/controllers/_array_viewer.py
```diff
@@ -163,6 +175,24 @@ def data(self, data: Any) -> None:
             self._data_model.data_wrapper = DataWrapper.create(data)
         self._fully_synchronize_view()
 
+    @property
+    def roi(self) -> RectangularROIModel | None:
+        """Return ROI being displayed."""
+        return self._roi_model
+
+    @roi.setter
+    def roi(self, roi_model: RectangularROIModel | None) -> None:
```
tlambert03, line 184: would be nice if this could accept the tuple bounds directly as well (rather than needing to import the object)

src/ndv/controllers/_array_viewer.py
```diff
@@ -163,6 +175,24 @@ def data(self, data: Any) -> None:
...
+        """Set ROI being displayed."""
+        # Disconnect old model
+        if self._roi_model is not None:
+            self._set_roi_model_connected(self._roi_model, False)
+
+        # Connect new model
+        self._roi_model = roi_model
+        if self._roi_model is not None:
+            self._set_roi_model_connected(self._roi_model)
+        self._fully_synchronize_view()
```
tlambert03, line 194: i think we might be using this method a bit too much.  it's something to look at in the future, but for the case of just setting the ROI, do we really need to throw away and rebuild all of the existing image textures?

src/ndv/controllers/_array_viewer.py
```diff
@@ -288,6 +343,36 @@ def _on_model_channel_mode_changed(self, mode: ChannelMode) -> None:
         self._clear_canvas()
         self._request_data()
 
+    def _on_roi_model_bounding_box_changed(
+        self, bb: tuple[tuple[float, float], tuple[float, float]]
+    ) -> None:
+        self._roi_view.set_bounding_box(*bb)  # type: ignore
```
tlambert03, line 349: why was the type ignore needed here?  seems correct

tlambert03: ahh, cause `_roi_view` might be None?  if there's absolutely no chance it will be None, let's use `cast()` (or use a more narrow `type: ignore [code]`, otherwise let's check `is not None`

src/ndv/controllers/_array_viewer.py
```diff
@@ -288,6 +343,36 @@ def _on_model_channel_mode_changed(self, mode: ChannelMode) -> None:
...
+
+    def _on_roi_model_visible_changed(self, visible: bool) -> None:
+        self._roi_view.set_visible(visible)  # type: ignore
```
tlambert03, line 352: same here with the type ignore

src/ndv/controllers/_array_viewer.py
```diff
@@ -288,6 +343,36 @@ def _on_model_channel_mode_changed(self, mode: ChannelMode) -> None:
...
+
+    def _on_interaction_mode_changed(self, mode: InteractionMode) -> None:
+        if mode == InteractionMode.CREATE_ROI:
+            # Create ROI model if needed to store ROI state
+            if self.roi is None:
+                self.roi = RectangularROIModel(visible=False)
```
tlambert03, line 358: any reason we shouldn't just create one in the init?  It's not too expensive right?

src/ndv/v1/_old_viewer.py
tlambert03, line 1: fwiw, I don't care whether we keep *updating* v1.  my intention was just to leave it as is in a little isolated namespace... not to keep updating it as we go

src/ndv/views/_jupyter/_app.py
```diff
@@ -52,19 +70,29 @@ def filter_mouse_events(
...
             if etype == "pointer_move":
-                mme = MouseMoveEvent(x=ev["x"], y=ev["y"])
+                mme = MouseMoveEvent(x=ev["x"], y=ev["y"], btn=active_btn)
                 receiver.on_mouse_move(mme)
+                if cursor := receiver.get_cursor(mme):
+                    canvas.cursor = cursor.to_jupyter()
                 receiver.mouseMoved.emit(mme)
             elif etype == "pointer_down":
-                mpe = MousePressEvent(x=ev["x"], y=ev["y"])
+                active_btn = btn
```
tlambert03, line 89: btn is technically `MouseButton | Unbound` here.  Since this is the only place you actually use it, perhaps move the `JupyterAppWrap.mouse_btn(ev["button"])` call here

src/ndv/views/_jupyter/_app.py
```diff
@@ -52,19 +70,29 @@ def filter_mouse_events(
         # patch the handle_event from _jupyter_rfb.CanvasBackend
         # to intercept various mouse events.
         super_handle_event = canvas.handle_event
+        active_btn: MouseButton = MouseButton.NONE
 
         def handle_event(self: RemoteFrameBuffer, ev: dict) -> None:
+            nonlocal active_btn
+            nonlocal canvas
```
tlambert03, line 77: typically you only need `nonlocal`/`global` if you're only going to *mutate* the variable.  but canvas doesn't get mutated here.  Also, since `active_btn` isn't even used in the surrounding scope, you could probably get rid of the nonlocal altogether

src/ndv/views/_qt/_app.py
```diff
@@ -67,19 +73,35 @@ def array_view_class(self) -> type[ArrayView]:
...
 class MouseEventFilter(QObject):
-    def __init__(self, canvas: QObject, receiver: Mouseable):
+    def __init__(self, canvas: QWidget, receiver: Mouseable):
         super().__init__()
         self.canvas = canvas
         self.receiver = receiver
+        self.active_button = MouseButton.NONE
+
+    def mouse_btn(self, btn: Any) -> MouseButton:
+        from qtpy.QtCore import Qt
```
tlambert03, line 92: move import to top

src/ndv/views/_qt/_array_view.py
```diff
@@ -237,15 +251,20 @@ def create_sliders(self, coords: Mapping[Hashable, Sequence]) -> None:
         """Update sliders with the given coordinate ranges."""
         layout = cast("QFormLayout", self.layout())
         for axis, _coords in coords.items():
-            sld = QLabeledSlider(Qt.Orientation.Horizontal)
-            sld.valueChanged.connect(self.currentIndexChanged)
+            # Create a slider for axis if necessary
+            if axis not in self._sliders:
```
tlambert03, line 255: ah interesting.  did you find this cause of weird or excessive behavior?  Or just while looking closely at the code?

tlambert03 (commented):
i think this is looking very nice, and the runtime behavior is great!  I looked most closely at everything outside of the actual vispy and pygx implementations (the general model and hookup of it all).  If there was anything about the specific graphics implementations that you want me to take a close look at, go ahead and comment inline.

otherwise, do whatever final cleanup/testing you want to add and then let's get this in!

src/ndv/_types.py
```diff
@@ -101,9 +106,46 @@ def to_qt(self) -> Qt.CursorShape:
...
+            CursorType.DEFAULT: "default",
+            CursorType.CROSS: "crosshair",
+            CursorType.V_ARROW: "ns-resize",
+            CursorType.H_ARROW: "ew-resize",
+            CursorType.ALL_ARROW: "move",
+            CursorType.BDIAG_ARROW: "nesw-resize",
+            CursorType.FDIAG_ARROW: "nwse-resize",
+        }[self]
+
+    @lru_cache
```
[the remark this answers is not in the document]
gselzer: Sure!

src/ndv/models/_roi_model.py
gselzer: Sure, I was thinking that a 3D bounding box could be useful. That could be another PR though - for now I'll just make it a `tuple[tuple[float, float], tuple[float, float]]`

src/ndv/controllers/_array_viewer.py
gselzer: We can, I had thought about doing it before. The reason why I hadn't is because if/when we want other rois, the tuple bounds will become ambiguous (i.e. are they the bounding box of a rectangle, or the endpoints of a line?), and I didn't want to implement something I thought of as valuable in the short-term.

Thoughts?

src/ndv/controllers/_array_viewer.py
gselzer: > i think we might be using this method a bit too much.

I agree! Until/unless we do something for ROIs like we did in #87, maybe we can just synchronize it in a separate method...

src/ndv/controllers/_array_viewer.py
gselzer: Right, yeah, don't want it to be `None`. Added the `is not None` check, not sure why I didn't do that before...

src/ndv/controllers/_array_viewer.py
gselzer: It's not expensive at all, why do you want to create one there?

src/ndv/views/_jupyter/_app.py
gselzer: > Since this is the only place you actually use it, perhaps move the `JupyterAppWrap.mouse_btn(ev["button"])` call here

Good point - moved!

> btn is technically `MouseButton | Unbound` here.

Why?

src/ndv/views/_jupyter/_app.py
gselzer: You're right - we don't need `nonlocal canvas`. Good catch!

I do think that we need `nonlocal active_btn` though - the mouse button is not provided on a mouse move event, so we need to maintain the currently-pressed button as state.

src/ndv/views/_qt/_app.py
gselzer: Good catch - probably leftover from when these were all in one class!

src/ndv/views/_qt/_array_view.py
gselzer: Yup, I found that two sets of sliders were created when you called `ArrayViewer._fully_synchronize_view` twice (which happened when we were calling that method in the `roi` setter).

src/ndv/views/_jupyter/_app.py
tlambert03: in this code:

```python
    etype = ev["event_type"]
    if "button" in ev:
        btn = JupyterAppWrap.mouse_btn(ev["button"])
    if etype == "pointer_move":
        ...
    elif etype == "pointer_down":
        active_btn = btn
```

you are implicitly equating `"button" in ev` with `etype == "pointer_down"`.  even if you checked and every time you ever saw `"button" in ev` you *also* say `etype == "pointer_down"` the code itself doesn't guarantee that (from a typing perspective), so while it's probably "safe", it's technically open to an Unbound error (for example if the upstream code ever happened to change that implicit relationship between `ev["event_type"]` and `ev["button"]`

src/ndv/views/_jupyter/_app.py
tlambert03: in this code:

```python
    etype = ev["event_type"]
    if "button" in ev:
        btn = JupyterAppWrap.mouse_btn(ev["button"])
    if etype == "pointer_move":
        ...
    elif etype == "pointer_down":
        active_btn = btn
```

you are implicitly equating `"button" in ev` with `etype == "pointer_down"`.  even if you checked and every time you ever saw `"button" in ev` you *also* say `etype == "pointer_down"` the code itself doesn't guarantee that (from a typing perspective), so while it's probably "safe", it's technically open to an Unbound error (for example if the upstream code ever happened to change that implicit relationship between `ev["event_type"]` and `ev["button"]`

src/ndv/views/_jupyter/_app.py
tlambert03: ahh, sorry, you're right.  because `canvas.handle_event = MethodType(handle_event, canvas)` maintains that closure for a (possibly) extended amount of time

src/ndv/controllers/_array_viewer.py
tlambert03: my bad: i had missed the fact that _roi_model can sometimes be *set* to `None` in the setter... so I mistakenly thought that this was just a lazy way to create a `RectangularROIModel`, but that there was not real need to have `self.roi` be `Optional`.  this is fine

src/ndv/views/_jupyter/_app.py
gselzer: Aha, good point! Should be fixed now.

src/ndv/controllers/_array_viewer.py
tlambert03: i'm somewhat neutral.  If you think that all ROI models will ultimately still be a sequence of 2-tuples, it feels like it won't be hard to maintain backwards compatibility if you accept `Sequence[tuple[float, float], ...]` and do runtime checks?  your call though

gselzer: > If you think that all ROI models will ultimately still be a sequence of 2-tuples, it feels like it won't be hard to maintain backwards compatibility if you accept `Sequence[tuple[float, float], ...]` and do runtime checks?

Well, in a perfect world I think 3D ROIs would be awesome, so if we did implement it I'd probably want `Sequence[Sequence[float]]` or some 2/3 tuple type alias.

> your call though

Since I'm envisioning further PRs anyways to revise this system (enabling multiple ROIs, different shapes, integrations with e.g. histograms, etc.) let's leave the API strict for now and let the eventual needs inform future design?

src/ndv/models/_roi_model.py
```diff
@@ -0,0 +1,40 @@
...
+    visible: bool = True
+    bounding_box: tuple[tuple[float, float], tuple[float, float]] = ((0, 0), (0, 0))
+
+    @field_validator("bounding_box", mode="after")
+    @classmethod
+    def _validate_bounding_box(
+        cls, bb: Any
+    ) -> tuple[tuple[float, float], tuple[float, float]]:
+        if not isinstance(bb, tuple):
+            raise ValueError(f"{bb} not a tuple of points!")
```
tlambert03, line 35:
```suggestion
        cls, bb: tuple[tuple[float, float], tuple[float, float]]
    ) -> tuple[tuple[float, float], tuple[float, float]]:
```

(because mode='after')

tlambert03 (approved):
LGMT! 🚀 automerge is on

PR #123, opened 2025-01-28
docs: Add a set of examples in `cookbook`

jacopoabramo:
A PR to provide a `cookbook` section in the documentation which can be used as a reference to use ndv in different contexts.

I added 3 examples on how to embed `ArrayViewer.widget()` in a custom Qt application (was also a learning experience for me!)

- `ndv` with a couple of pushbuttons to select one dataset or the other;
- multiple `ndv.ArrayViewer` widgets in the same `QWidget`;
- a pseudo-microscope dashboard using `openwfs`.

In the first example I also wanted to add a button to show the `cells3d` dataset but after loading it, the 2D datasets were squashed and I did not know how to fix the behavior; for a minimal example though I believe it should work even without `cells3d`


docs/cookbook.md
```diff
@@ -0,0 +1,332 @@
+# `ndv` cookbook
+
+This section of the documentation provides some useful recipes to use `ndv` in different contexts,
+providing an example application for each possible use case.
+
+## Embed `ArrayViewer` in a `QWidget`
+
+`ndv` can be embedded in an existing Qt application and enriched with additional elements in a custom layout.
```
tlambert03, line 8: let's make an entire cookbook folder, with individual files for recipes.

So 
- create docs/cookbooks
- rename this file to docs/cookbooks/embedding.md
- remove the `ndv` cookbook header, and decrease all of headers by one level.

docs/cookbook.md
```diff
@@ -0,0 +1,332 @@
...
+
+    def _load_astronaut(self) -> None:
+        self._viewer.data = None
+        self._viewer.data = astronaut()[:, :, 0]
+
+app = QtWidgets.QApplication([])
+widget = EmbeddingWidget()
+widget.show()
+run_app()
+```
```
tlambert03, line 74: we can avoid the duplication here using [snippets](https://facelessuser.github.io/pymdown-extensions/extensions/snippets/)

`````suggestion
````python title="examples/cookbook/ndv_embedded.py"
--8<-- "examples/cookbook/ndv_embedded.py"
````
`````

File: docs/cookbook.md [hunk 3]
```diff
@@ -0,0 +1,332 @@
...
+        layout.addWidget(self._cells_virewer.widget(), 0, 5, 4, 4)
+
+        self.setLayout(layout)
+
+
+app = QtWidgets.QApplication([])
+widget = MultiNDVWrapper()
+widget.show()
+run_app()
+```
```
tlambert03, line 123:
`````suggestion
````python title="examples/cookbook/multi_ndv.py"
--8<-- "examples/cookbook/multi_ndv.py"
`````

File: docs/cookbook.md [hunk 4]
```diff
@@ -0,0 +1,332 @@
...
+manager = SyntheticManager()
+wrapper = DashboardWidget()
+
+manager.sigNewFrame.connect(wrapper.new_frame)
+wrapper.sigSimStarted.connect(manager.toggle_simulation)
+wrapper.sigMoveStage.connect(manager.move_stage)
+
+wrapper.show()
+app.exec()
+```
```
tlambert03, line 330:
`````suggestion
````python title="examples/cookbook/microscope_dashboard.py"
--8<-- "examples/cookbook/microscope_dashboard.py"
````
`````

File: docs/screenshots/multi_ndv.png
Comment (tlambert03) [inline docs/screenshots/multi_ndv.png:1]: these are all fantastic, and i definitely want to be able to show them.  However, from past experience, I'm somewhat hesitant to commit them to the repo due to size bloat.  This file alone is 0.5 MB, and this sort of pattern can quickly grow the size of a repo (and even when removed in the future, you have to essentially rewrite git history in order to get your repo size back down).  

I would much rather generate these at documentation build time.  Yes, that adds complexity, but it avoids repo bloat, and also ensures that the scripts are actually still working with the current API.  I'm happy to work on that.  I could either do it as a PR to this PR, or as a followup PR (but we would temporarily omit the screenshots from this PR)

File: mkdocs.yml [hunk 5]
```diff
@@ -18,6 +18,7 @@ nav:
   - Home: index.md
   - install.md
   - motivation.md
+  - cookbook.md
   - env_var.md
```
tlambert03, line 22: after the suggestion above, this would then become:

```suggestion
  - env_var.md
  - Cookbook: cookbook/
```

(note, I moved it below env_var ... because it looks a bit better once it becomes a folder)

tlambert03 (commented):
super cool!  ❤️  thank you!

I have a few suggestions/requests.  I'll mark them inline

docs/screenshots/multi_ndv.png
jacopoabramo: Yes I also had a feeling that it would have been a bad idea; I committed the PNGs nonetheless hoping for some feedback.

As a startup I can remove the images and as you suggested add them in a follow-up PR. I don't think its critical to show them at this time, but its definetely a nice "marketing move"

tlambert03: absolutely :) they make the page look *wayyyy* better

examples/cookbook/microscope_dashboard.py
```diff
@@ -0,0 +1,190 @@
+"""A minimal microscope dashboard.
```
tlambert03, line 1: by the way, it doesn't necessarily have to happen in this PR, but a really nice way to express additional dependencies in an example script is by using [PEP 722](https://peps.python.org/pep-0722/).  Somewhat amazingly, you can then just run that script directly with [uv run](https://docs.astral.sh/uv/guides/scripts/#declaring-script-dependencies), entirely free of a python environment: `uv run ./examples/cookbook/microscope_dashboard.py`

```suggestion
# /// script
# dependencies = [
#   "astropy",
#   "openwfs",
#   "ndv[vispy,pyqt]",
# ]
# ///
"""A minimal microscope dashboard.
```

jacopoabramo: That's nifty! It works only with `uv` I'm guessing? But still since you're using it with `ndv` I guess it makes sense and it's not a big effort to add.

jacopoabramo:
Hopefully that should do it

examples/cookbook/microscope_dashboard.py
tlambert03: yeah, it's a general guideline.  (and apologies, I said PEP 722, but it's actually 723 ... 722 was rejected: https://peps.python.org/pep-0723/)

uv is just early to the game (as usual) of supporting the general pep in a cool way.  You can also use pipx: https://pipx.pypa.io/stable/examples/#pipx-run-examples

PR #126, opened 2025-01-28
Add mechanism to generate screenshot from examples, and build during docs

tlambert03:
follow up on #123  (cc @jacopoabramo), this adds a mechanism to run a script, and capture the output to a buffer that gets written to a temporary directory used by mkdocs.  allowing us to embed screenshots based on included examples in the documentation.

still some Qt bugs to iron out (which may require modificiations to the scripts themselves)

### DIFF
--- a/docs/cookbook/embedding.md
+++ b/docs/cookbook/embedding.md
@@ -11,7 +11,7 @@ The following script shows an example on how to dynamically select a data set an
 --8<-- "examples/cookbook/ndv_embedded.py"
 ````
 
-![Screenshot generated from ndv_embedded.py](../screenshots/cookbook/ndv_embedded.png)
+![Screenshot generated from ndv_embedded.py](../screenshots/cookbook/ndv_embedded.png) {: .screenshot }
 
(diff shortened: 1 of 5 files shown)


jacopoabramo:
Nice! I didn't know you could do this with mkdocs. I wonder if sphinx has something similar

tlambert03:
oh for sure.

tlambert03:
ok, i did some more work here, significantly cleaned up the pattern, directly using mkdocs API instead of the mkdocs-gen-files plugin.

@jacopoabramo, I modified your scripts a bit to follow patterns i would "suggest", that was in https://github.com/pyapp-kit/ndv/pull/126/commits/9d27af1a4199045cb18e576fdc8ed5c75dac1ea2  ... with the biggest change being to the stage widget in the dashboard. hope you don't mind.

also... this now also generates both light and dark themed images that change with the docs theme being used! 🎉

Screenshot 2025-01-29 at 9 54 35 AM
Screenshot 2025-01-29 at 9 54 40 AM

tlambert03:
yep, I had the same thought 👍

PR #132, opened 2025-02-07
feat: per-channel histograms

gselzer:
This PR aims to provide additional functionality to the histograms:
* The single histogram button has been removed in favor of histogram buttons on the widget LUT view, allowing multiple histograms to be displayed at once:

https://github.com/user-attachments/assets/157df6c9-0e61-413e-bf66-f49d365606ba

* Support for drawing histograms has been added to the wx frontend (and planned for jupyter as well).

https://github.com/user-attachments/assets/fd0458be-f4f7-4cb8-aa21-701a1ce4c811

It's not ready for code review yet, but filing a PR now for feedback on the UI aspects of things!

TODO:
* [x] Jupyter UI controls for histograms
* [x] Squashing wx bugs (e.g. needing to drag the window to get the histogram to display properly)
* [x] Purge TODOs/clean - the changes are ugly right now
* [ ] Test?


tlambert03:
loooove the UI in Qt. 

as a parallel option, I imagine that we may also want to overlay the different channel histograms on the same canvas (to save space).  But since that makes clim/gamma control much harder, this is definitely something we'll also want.  I like how it's next to the auto button too

gselzer:
@tlambert03 I still want to increase test coverage a little more here, however I've gotten this code cleaned up to the point where I think it's ready for other eyes. Let me know what you think of the design!

src/ndv/views/_qt/_array_view.py
```diff
@@ -158,20 +159,41 @@ def __init__(self, parent: QWidget | None = None) -> None:
...
+        top.setSpacing(5)
+        top.setContentsMargins(0, 0, 0, 0)
+        top.addWidget(self.visible)
+        top.addWidget(self.cmap)
+        top.addWidget(self.clims)
+        top.addWidget(self.auto_clim)
+        top.addWidget(self.histogram_btn)
+
+        # TODO: Replace with histogram
+        self._histogram_container = QHBoxLayout()
```
tlambert03, line 177: I would just make this a widget rather than a layout.  it's much more natural to toggle visibility of a widget than it is to iterate through all items in a layout and toggle their visibility.  (down in `_on_q_histogram_toggled`)

src/ndv/views/_qt/_array_view.py
```diff
@@ -158,20 +159,41 @@ def __init__(self, parent: QWidget | None = None) -> None:
...
-        layout.addWidget(self.clims)
-        layout.addWidget(self.auto_clim)
+        layout.addLayout(top)
+        layout.addLayout(self._histogram_container)
 
 
 class QLutView(LutView):
-    def __init__(self) -> None:
+    # NB: In practice this will be a ChannelKey but Unions not allowed here.
+    histogramRequested = psygnal.Signal(object)
```
tlambert03, line 189: too bad to see signals coming back here :) but I agree that it's the correct approach for now:  when we get around to doing something like you did in #87  to the ArrayView (i.e. directly mutating some model rather than emitting a signal), then I assume this will be removable again.

src/ndv/controllers/_array_viewer.py
tlambert03, line 1: everything in here looks great.  feels very natural

src/ndv/views/_jupyter/_array_view.py
```diff
@@ -347,16 +377,23 @@ def _on_channel_mode_changed(self, change: dict[str, Any]) -> None:
         """Emit signal when the channel mode changes."""
         self.channelModeChanged.emit(ChannelMode(change["new"]))
 
+    def add_histogram(self, channel: ChannelKey, widget: Any) -> None:
```
tlambert03, line 380: ❤️🎉

src/ndv/views/_qt/_array_view.py
```diff
@@ -158,20 +159,41 @@ def __init__(self, parent: QWidget | None = None) -> None:
...
+        top.addWidget(self.clims)
+        top.addWidget(self.auto_clim)
+        top.addWidget(self.histogram_btn)
+
+        # TODO: Replace with histogram
+        self._histogram_container = QHBoxLayout()
+        self._histogram_container.setSpacing(5)
+        self._histogram_container.setContentsMargins(0, 0, 0, 0)
+
+        layout = QVBoxLayout(self)
```
tlambert03, line 181: I think this probably needs `layout.setSpacing(0)`. keeps the histogram closer to its associated slider than the next one...

src/ndv/views/_wx/_array_view.py
```diff
@@ -51,6 +55,24 @@ def __init__(self, parent: wx.Window):
         self.SetSizer(sizer)
 
 
+def _add_icon(btn: wx.AnyButton, icon: str) -> None:
```
tlambert03, line 58: love this.  something still not working on my end (macos) but haven't looked closer:
Screenshot 2025-02-21 at 1 51 49 PM

tlambert03 (commented):
i love this!  very clean.   let me know when you're ready for merge.  I'll take a final pass at that point, but I don't see anything in here i don't like

src/ndv/views/_qt/_array_view.py
gselzer: Sure, removed.

src/ndv/views/_qt/_array_view.py
gselzer: Oop, good catch!

gselzer: Fixed!

src/ndv/views/_wx/_array_view.py
gselzer: Huh, that's unfortunate, and it's interesting that the "Reset Zoom" icon does show...I'd be interested if you could debug. Also happy to not use the icons on that frontend for the time being

image

gselzer:
> i love this! very clean. let me know when you're ready for merge.

I think it's ready for merge once we figure out the icons!

tlambert03:
did some small changes in 6de995d ... First, I noticed that without setting a `label`, the button lost its height (which I think was the problem before).  but then even better I found `SetBitmapLabel` which has the desired effect of replacing the label if one is present, and retaining the dimensions.

Screenshot 2025-02-24 at 1 32 44 PM

gselzer:
> but then even better I found `SetBitmapLabel` which has the desired effect of replacing the label if one is present, and retaining the dimensions.

Aha! Looks great! You're happy with the optional dependency on `pyconify`? (Assuming it's an implicit dependency with qt anyways)

PR #136, opened 2025-02-12
feat: add `call_later` (public) and `process_events` (private) to the NDVApp API

tlambert03:
Something we generally need to be able to do in all gui frontends is call the equivalent of Qt's `QTimer.singleShot()` or `QApplication.processEvents()`.  this adds those two things to the App protocol, and exposes `ndv.call_later` (which is convenient for examples, and such)


gselzer:
Curious as to your desire for `process_events` if it is (a) private and (b) unused within the repo?

tlambert03:
It is used in streaming viewer... however there too I ended up moving it to the example.  So it does need considering

PR #137, opened 2025-02-12
test: skip ubuntu-vispy-pyside-3.12

tlambert03:
this particular combination always fails on CI.


gselzer:
Are you sure that this one always segfaults? Most of my CI jobs on #114 (which are up-to-date with `main`) show that job passing ([example](https://github.com/pyapp-kit/ndv/actions/runs/13293671121))

tlambert03:
It's definitely not the only one, and it's not _always_ seg faulting either .  I thought it was the main culprit when I was doing my testing? But no, not sure

PR #139, opened 2025-02-14
feat: add widget visibility options to ArrayViewerModel

tlambert03:
In #134, I want to be able to reuse the ArrayView objects, but hide some of the widgets that aren't relevant in that context (such as the 3D button, and other widgets).  This PR adds various options to `ArrayViewerModel` that control the visibility of various elements:

```
    show_3d_button: bool = True
    show_histogram_button: bool = True
    show_reset_zoom_button: bool = True
    show_roi_button: bool = True
    show_channel_mode_selector: bool = True
    show_progress_spinner: bool = False
```

### DIFF
--- a/src/ndv/views/bases/_array_view.py
+++ b/src/ndv/views/bases/_array_view.py
@@ -74,7 +74,3 @@ class ArrayView(Viewable):
 
     def remove_histogram(self, widget: Any) -> None:
         raise NotImplementedError
-
-    # could be moved to set_options?
-    def set_progress_spinner_visible(self, visible: bool) -> None:
-        return
--- a/src/ndv/views/bases/_array_view.py
+++ b/src/ndv/views/bases/_array_view.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from abc import abstractmethod
 from collections.abc import Sequence
+from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any
 
 from psygnal import Signal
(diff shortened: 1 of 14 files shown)


src/ndv/views/bases/_array_view.py
```diff
@@ -19,6 +20,17 @@
...
 
+@dataclass
+class ArrayViewOptions:
+    """Options for configuring an ArrayView widget."""
+
+    show_3d_button: bool | None = None
+    show_histogram_button: bool | None = None
+    show_reset_zoom_button: bool | None = None
+    show_roi_button: bool | None = None
+    show_channel_mode_selector: bool | None = None
```
gselzer, line 31: I almost wonder whether this might be a natural fit for the `ArrayViewerModel`...thoughts?

gselzer (commented):
I like the functionality here, and have also desiring similar functionality for disabling/hiding some of the controls. For example, in #132 I've been thinking that it would be nice to disable the per-channel histograms, and it seems that would be another possible option here.

More broadly I've also been wondering (and it's only loosely related to this PR, I'm just writing out thoughts) whether the mechanism of creating viewers could be refined (maybe a declaritive, builder-style?) for more flexibility and modularity (i.e. making it easier to create custom viewer layouts, enabling/disabling different features, etc.).

src/ndv/views/bases/_array_view.py
tlambert03: yeah, lemme try.  that would also be the natural place to define a more declarative spec for viewer widget construction.

tlambert03: changed.  i much prefer it.  thanks for the suggestion.

Issue #131, opened 2025-02-06
bug: Dropdowns incorrectly positioned on PyGFX + Qt


gselzer:
For some reason, the Qt dropdowns are translated across the screen when I'm using PyGFX canvases (in the video, they're in the bottom right corner):

https://github.com/user-attachments/assets/0b8f95bf-b72a-44a3-8ffe-df1ca9a8ff77

For some reason, it does not happen if I instead use VisPy.

For what it's worth, I'm using PyQt6 v6.8.0.

@tlambert03 can you replicate?

The only hint I have here is that when I run PyGFX, I see logged to the terminal
```
QWindowsWindow::setGeometry: Unable to set geometry 600x666+0+0 (frame: 600x666+0+0) on QWidgetWindow/"QWidgetClassWindow" on "DELL U2713H". Resulting geometry: 600x666+2160+1072 (frame: 600x666+2160+1072) margins: 0, 0, 0, 0)
```
So maybe it pertains to my monitor? If I launch from the other monitor, I see something different (this bug I've actually noticed before, but now it's more annoying):

https://github.com/user-attachments/assets/842ed489-08ad-4705-8c60-1743cfc5e09d

Maybe it's a bug in the pygfx qt backend (cc @almarklein)? Maybe this would be fixed with [rendercanvas](https://github.com/pygfx/rendercanvas)? I'll have to debug more...

tlambert03:
Using render canvas elsewhere… need to update here as well 👍

tlambert03:
@fdrgsp ... here's the issue we observed yesterday

PR #140, opened 2025-02-15
refactor: use rendercanvas instead of wgpu.gui

tlambert03:
pygfx is moving to using rendercanvas.  This does it too.

there's still something weird going on with wx (segfaults, and issues with timers).  need to troubleshoot that

### DIFF
--- a/src/ndv/data.py
+++ b/src/ndv/data.py
@@ -50,7 +50,7 @@ def nd_sine_wave(
                 # Assign to the output array
                 output[angle_idx, freq_idx, phase_idx] = sine_wave
 
-    return output
+    return output.astype(np.float32)
 
 
 def cells3d() -> np.ndarray:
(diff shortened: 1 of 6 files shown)


gselzer:
Hey, great!

I was very close to filing this PR myself a couple days ago. Just an FYI that I've been putting some work into the wx backend in #134 but it's likely tangential to the fixes you have here!

gselzer:
Whoops

tlambert03:
unfortunately no, i checked on windows and this doesn't solve #131.   I think that is may well be related to the `winID()` call... which still, at the moment, seems like it might be a big issue for using pygfx.  it needs some (deep, os-specific, sleuthing).

https://github.com/pygfx/rendercanvas/issues/55

tlambert03:
while this doesn't fix much, it's still an improvement and can go in as is

PR #141, opened 2025-02-20
feat: rgb, v2

gselzer:
This PR revives the work done in #41, but updated for the new viewer. I'm guessing that the design can be further refined, and there are more bugs to be ironed out.

image

Still lots of cleaning, testing, bugfixing to do. Notably, there were many comments on #41 that should be gone through and addressed here.


src/ndv/data.py
```diff
@@ -144,3 +145,16 @@ def cosem_dataset(
...
+    img = np.zeros((4, 256, 256), dtype=np.uint8)
+
+    for x in range(256):
+        for y in range(256):
+            img[0, x, y] = x
+            img[1, x, y] = y
+            img[2, x, y] = 255 - x
+            img[3, x, y] = int(sqrt((x - 128) ** 2 + (y - 128) ** 2))
+
+    return img
```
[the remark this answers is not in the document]
tlambert03: I'd be happy only using astronaut, but if you find this one useful for debugging, then I don't mind leaving it in (but add to the docstring when someone might want to use it)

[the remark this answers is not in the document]
gselzer: Hmm, well I just wrote a commit deleting it, however I'm now realizing that there's no existing dataset with an alpha channel...so I added it back in 3761b9b, but augmented it into a 3D RGB dataset (with a short docstring).

gselzer:
@tlambert03 I'm generally happy with the functionality here and would appreciate your thoughts and high-level review. Couple points I'd like your opinions on:
* I've changed the `LutKey` type alias into `ChannelKey`, which now is basically anything `Hashable`. The reason that I need this is because I need a specific identifier for RGB mode - it cannot be an integer, and must be differentiable from `None` - to trigger an appropriate RGB `LutView` and to correctly slice (read: *not* slice) the data. However this typing may be too broad - what do you think?
* On a broader level, I'm not quite satisfied with how much RGB mode gets special-cased - in a sense, it needs special casing because it's the only mode we (want to) support where the channels must be combined for meaning - but it would be nice to minimize the special cases.
* The histogram right now is lacking because there's no special RGB support - I'm thinking this may best be solved in another PR, because it's not technically wrong right now.
* There's an obnoxious case where, if you do not specify RGB mode on viewer construction (i.e. remove the data model from being passed to `imshow` in the new example, you'll start out in composite mode where the channel is one of the visible axes. I try to make switching to RGB do the right thing and change the visible axes (in `_ArrayDataDisplayModel._on_channel_mode_change`), but then there's no way to go back to the previously shown axes. This feels unfortunate/wrong (hence the leftover `FIXME`), but maybe it's just necessary?

gselzer:
@tlambert03 did you ever get a chance to take a look at this?

src/ndv/models/_data_display_model.py
```diff
@@ -245,27 +262,26 @@ def request_sliced_data(
...
         vis_ax = req.visible_axes
         t_dims = vis_ax + tuple(i for i in range(data.ndim) if i not in vis_ax)
 
-        if (ch_ax := req.channel_axis) is not None:
-            ch_indices: Iterable[Optional[int]] = range(data.shape[ch_ax])
+        data_response: dict[ChannelKey, np.ndarray] = {}
+        ch_ax = req.channel_axis
+        # For RGB and Grayscale - keep the whole array together
+        if self.display.channel_mode == ChannelMode.RGBA:
+            data_response["RGB"] = data.transpose(*t_dims).squeeze()
```
tlambert03, line 277: one reason for making this a static method was that i imagine this method coming off of this class eventually, to a global executor that receives data requests and returns data responses.  So, ideally, the `DataRequest` itself should contain *all* of the information needed to return a response (there should be no external state required, like `self.display.channel_mode`.  Id rather change the request object itself if needed.  Can we do that here?

src/ndv/views/bases/_array_view.py
```diff
@@ -65,7 +65,7 @@ def hide_sliders(
         self, axes_to_hide: Container[Hashable], *, show_remainder: bool = ...
     ) -> None: ...
     @abstractmethod
-    def add_lut_view(self, channel: ChannelKey) -> LutView: ...
+    def add_lut_view(self, key: ChannelKey = None) -> LutView: ...
```
tlambert03, line 68: do we call this anywhere with no arguments?

src/ndv/views/_vispy/_array_canvas.py
```diff
@@ -52,7 +52,7 @@ def data(self) -> np.ndarray:
             return self._visual._last_data  # type: ignore [no-any-return]
 
     def set_data(self, data: np.ndarray) -> None:
-        if not data.ndim == self._ndim:
+        if data.ndim not in {2, 3}:
```
tlambert03, line 55: this does now leave open the possibility of erroneously trying to set a Volume with 2d data right?

examples/rgb.py
```diff
@@ -0,0 +1,5 @@
+import ndv
+from ndv.models import ArrayDisplayModel
+
+display = ArrayDisplayModel(channel_mode="rgb")
```
tlambert03, line 4: I do think a little magic is acceptable here.  if someone uses `imshow(data.astronaut())`, it should look better than it currently does.  (this is also true on `main` for what it's worth... where it looks like we're not yet seeing that a YXC image has channel_axis 2).  So, if you want to add a little logic that picks "channel_mode=rgb" if someone hasn't explicitly set the channel mode or channel axis, and if the last axis has length 3 or 4

tlambert03 (approved):
beyond the few small comments, i think everything in here looks good!  nice job adding a  new feature with minimal lines :)

src/ndv/models/_data_display_model.py
gselzer: Good point - reverted to a static method in 4f388c3

src/ndv/views/bases/_array_view.py
gselzer: ...not anymore! I think I had a default in early on in this effort, but removed it in 0b688d5

src/ndv/views/_vispy/_array_canvas.py
gselzer: Uh, yes, probably. Fixed in 41479b0

examples/rgb.py
gselzer: Let me know what you think about the magic in 5ce2960 (plus the fix in 54b9a55) - took a little more massaging than I hoped but seems to work!

tlambert03:
remaining segfaults do seem to be specific to this PR.  locally, I can run `make test extras=pyqt,pygfx py=3.10` over and over on main with no issue, and this PR is giving segfault each time.  Will try to dig

tlambert03:
when running with lldb, I get this:

```
tests/test_examples.py::test_example[rgb.py] Process 56522 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x27c)
    frame #0: 0x0000000129340fa4 QtWidgets`QWidgetPrivate::drawWidget(QPaintDevice*, QRegion const&, QPoint const&, QFlags<QWidgetPrivate::DrawWidgetFlag>, QPainter*, QWidgetRepaintManager*) + 3152
QtWidgets`QWidgetPrivate::drawWidget:
->  0x129340fa4 <+3152>: ldr    w9, [x8, #0x228]
    0x129340fa8 <+3156>: tbnz   w9, #0x1e, 0x129340fd4 ; <+3200>
    0x129340fac <+3160>: ldrh   w8, [x23, #0x18]
    0x129340fb0 <+3164>: cbnz   w8, 0x129340fe4 ; <+3216>
(lldb)  bt
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x27c)
  * frame #0: 0x0000000129340fa4 QtWidgets`QWidgetPrivate::drawWidget(QPaintDevice*, QRegion const&, QPoint const&, QFlags<QWidgetPrivate::DrawWidgetFlag>, QPainter*, QWidgetRepaintManager*) + 3152
    frame #1: 0x000000012933ffe0 QtWidgets`QWidgetPrivate::paintOnScreen(QRegion const&) + 404
    frame #2: 0x000000012933fdb8 QtWidgets`QWidgetPrivate::syncBackingStore() + 196
    frame #3: 0x00000001293512f0 QtWidgets`QWidget::event(QEvent*) + 1420
    frame #4: 0x0000000128ab1160 QtWidgets.abi3.so`sipQWidget::event(QEvent*) + 224
    frame #5: 0x0000000129300298 QtWidgets`QApplicationPrivate::notify_helper(QObject*, QEvent*) + 336
    frame #6: 0x0000000129301d18 QtWidgets`QApplication::notify(QObject*, QEvent*) + 3260
    frame #7: 0x0000000128a995d0 QtWidgets.abi3.so`sipQApplication::notify(QObject*, QEvent*) + 248
    frame #8: 0x000000012845e390 QtCore`QCoreApplication::notifyInternal2(QObject*, QEvent*) + 200
    frame #9: 0x000000012845f49c QtCore`QCoreApplicationPrivate::sendPostedEvents(QObject*, int, QThreadData*) + 532
    frame #10: 0x000000012e86914c libqcocoa.dylib`___lldb_unnamed_symbol1725 + 360
    frame #11: 0x000000012e867800 libqcocoa.dylib`___lldb_unnamed_symbol1718 + 1008
```


so there's something going on during the paint event

tlambert03:
wow... i think i might have fixed it with [this change](https://github.com/pyapp-kit/ndv/pull/141/commits/7929362a6f7b8789a73433a53e42a99c85739b41#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128R64).  What I think is happening is that running `examples/rgb.py`, on pygfx, hits the warning in the LutView:

```
/Users/talley/dev/self/ndv/src/ndv/views/bases/_lut_view.py:93: UserWarning: Cannot set colormap on an RGB image
  self.set_colormap(model.cmap)
```

I believe that warning was getting upcast in tests to an exception.  For some reason, simply letting the application process events (where it would have called `exec()` is enough to let it fail gracefully).

However, it raises a new point.  @gselzer, it probably shouldn't show a warning to simply call `ndv.imshow(ndv.data.rgba())`, so perhaps that pygfx warning is either too much?  or we should make it a bit more specific such that it only warns if the user is somehow specifically trying to do something they shouldn't (which, i think might be a bit hard to detect at the moment).  I'm fine just removing the warning...

tlambert03:
one more thing, could we add a simple test (in addition to the example test) in `test_controller` that adds an RGB image, and asserts some basic things, such as that the magic is working.

gselzer:
> However, it raises a new point. @gselzer, it probably shouldn't show a warning to simply call `ndv.imshow(ndv.data.rgba())`, so perhaps that pygfx warning is either too much?

Yeah, it's probably too much - I'm a fan of silently ignoring changes to the colormap.

> one more thing, could we add a simple test (in addition to the example test) in `test_controller` that adds an RGB image, and asserts some basic things, such as that the magic is working.

Sure! I can give that a shot tomorrow!

gselzer:
@tlambert03 I've added a RGB magic test, and have removed the warning for setting the cmap on an RGB image. I additionally noticed that the wx/jupyter RGB lut views had a misplaced label (likely from the `main` merge), and fixed them.

This took a lot longer than I thought, as I was plagued by qt widget leaks. I'm fairly certain they're coming from the `_QLutView` itself (specifically from the cmap combo box), however I was not successful in pinpointing the issue. I'm pretty sure they arise anytime we create an `ArrayView` with data passed through the constructor (calling `ArrayViewer._fully_synchronize_view`), which is a pathway that we haven't tested explicitly without allowing leaks. Maybe something to look at in the future...

tlambert03:
> I'm fairly certain they're coming from the _QLutView itself (specifically from the cmap combo box), however I was not successful in pinpointing the issue. I'm pretty sure they would also arise anytime we create an ArrayView with data passed through the constructor (calling ArrayViewer._fully_synchronize_view), but we don't ever test this pathway without allowing leaks. Maybe something to look at in the future...

thanks, that's a great lead

src/ndv/controllers/_array_viewer.py
```diff
@@ -563,7 +598,15 @@ def _get_values_at_world_point(self, x: int, y: int) -> dict[ChannelKey, float]:
         values: dict[ChannelKey, float] = {}
         for key, ctrl in self._lut_controllers.items():
             if (value := ctrl.get_value_at_index((y, x))) is not None:
-                values[key] = value
+                # Handle RGB
+                if key == "RGB" and isinstance(value, np.ndarray):
```
tlambert03, line 602: just nothing for the future.  I don't love the addition of another special unique key here (I recognize that `None` was already that, and I don't have an immediate alternate suggestion at the moment).  absolutely fine for now, and maybe forever... just noticed it :)

tlambert03 (approved):
LGTM!

src/ndv/controllers/_array_viewer.py
gselzer: Yeah...I'm also not a fan, as I mentioned [here](https://github.com/pyapp-kit/ndv/pull/141#issuecomment-2675830703), but I wrote that for lack of a better option...very open to improving it later!

tlambert03: Ah yes sorry I forgot you pointed that out.  Agreed though, it’s a fine workaround

PR #142, opened 2025-02-20
Purge v1

gselzer:
This PR removes the original NDV application, and I believe that now is the right time to remove it because:
* With the merge of #114 we should now be at feature parity
* Tests specific to v1 are now failing in #141, meaning it will take active maintenance to maintain. While I haven't gotten to the bottom of these failures, I can't reproduce them easily either.

What do you think, @tlambert03?


tlambert03:
Kill it :)

PR #146, opened 2025-03-01
feat: Better histogram control (plus pygfx histogram overhaul)

gselzer:
This PR adds:
* A button to toggle the logarithmic base for the histogram. Originally, I was thinking that a slider would be nice (as I remembered @tlambert03 showing me some similar sort of functionality in Elements) but I couldn't get it to work nicely so for now the button will accomplish our goals.
* A button to reset the histogram view, because I got tired of losing it.
* A retained histogram frustum when setting the data. I find it frustrating when you pan/zoom to have a nice view of the data for that to be reset when you update the data.

To accomplish these features, some serious TLC was required to accomodate the new histogram features using pygfx. The way that we were calculating the margins was becoming increasingly complicated, so I developed a new strategy where the plot, x-axis, and y-axis are all rendered separately. This enables us to easily:
* Create one scene for canvas elements that should not move in response to user input, like the x-axis.
* Provide margins with far less computation.
The downside here is more complication in the rendering pipeline, but I think it's better than it was.

https://github.com/user-attachments/assets/ee2266fe-a63f-4772-94c7-1cc7ab096831

These controls should be ready for opinions (cc @tlambert03, @fdrgsp), but also still need to:
* [x] Resolve all `FIXME`s
* [x] Clean
* [ ] Test


tlambert03:
also great stuff here.  if implementing the slider was problematic, i'm fine with a button for now.  I was recently playing with histograms as well, and trying to apply range limits (so that the thing doesn't scroll past, for example, 0-65,535).  I think it's pretty straightforward using the 1DPanZoom camera, and updating the `rect.setter` so that you can never set the rect larger than the full range.  (we should probably only do it for integer dtypes though)

I'll have a closer look at the changes required here soon.  thanks!

gselzer:
> I was recently playing with histograms as well, and trying to apply range limits (so that the thing doesn't scroll past, for example, 0-65,535).

Let me know what you think of 5bf0f1e - could use a little polish but I think it works pretty well.

gselzer:
Particular review going to acedb6e would be worthwhile, as it adds a new field `clim_bounds` to the `LutModel`. The problem I was trying to solve here is "how do we notify each `LutView` (specifically the clims sliders and the histograms) about the maximum allowed clim?", which is both needed to solve this:

> I was recently playing with histograms as well, and trying to apply range limits (so that the thing doesn't scroll past, for example, 0-65,535)

as well as #152, which this PR would now close. However there may be a better place to store this state...

gselzer:
@tlambert03 I think this PR is probably ready for some eyes, in particular I'd like your feedback on:
* The addition of `LUTModel.clim_bounds`
* If you have any intuition on the pygfx histogram test qt leaks (see d142963)

tlambert03 (approved):
hey @gselzer,

i took a brief look at this in the time I just had available and I like everything I see.   I'm liking everything i see, and as usual: all your changes indicate a deep understanding of the system and they're all for the better.  since I'm a bit busy this week, if you want to just merge this at any point, feel free to, (or specifically ask anything else).   I'm happy with clim_bounds, and 5bf0f1e27cd7815494bf9a18b8f736d57adea06b looks fine to me. 

haven't looked too deeply at the pygfx changes. but the behavior seems good (outside of a sizing issue as shown in the screenshot below when using histogram in pygfx... i think it might need an explicit size policy?) 

Screenshot 2025-03-29 at 5 45 40 PM


I'll take a look at the leaks when I have a moment (but skipping them now is also ok with me)

tests/views/_pygfx/test_histogram.py
```diff
@@ -0,0 +1,110 @@
...
+def histogram() -> PyGFXHistogramCanvas:
+    canvas = PyGFXHistogramCanvas()
+    canvas.set_range(x=(0, 10), y=(0, 1))
+    return canvas
+
+
+# FIXME: These leaks are very consistent
+@pytest.mark.allow_leaks
+@pytest.mark.usefixtures("any_app")
+def test_interaction(model: LUTModel, histogram: PyGFXHistogramCanvas) -> None:
```
tlambert03, line 36: i think you can avoid the leak by not making this fixture.  particularly since the fixture isn't used anywhere besides a single test (and even if it was), I'd say just replace it with 

```python
histogram = PyGFXHistogramCanvas()
histogram.set_range(x=(0, 10), y=(0, 1))
```

it makes it slightly clearer whats going on (without having to look elsewhere for the fixture).  If the fixture starts to get very complex, and is reused in a lot of places... perhaps we can revisit, but I suspect it's holding on to the rendercanvas even after the test is finished

gselzer: Yes, this probably helps, however I think it was equally important to call `histogram.close()` at the end.

PR #147, opened 2025-03-01
feat: Add autoscale tail ignore via popup

gselzer:
This PR adds GUI functionality for autoscaling using `ClimsPercentile`, where the user can ignore the outlier samples of the dataset being displayed in autoscale:

https://github.com/user-attachments/assets/3c832b83-904c-400e-b4f7-58cde99a412d

I like the placement as a context menu because it means all autoscaling functionality appears within the same area of the canvas and because it normally does not take up any screen space.

Unfortunately, I do not know whether this design is possible with all our frontends, seeing as how I cannot figure out how to write a context menu in ipywidgets.

We might instead have to settle for adding this widget to the layouts added in #146...

### DIFF
--- a/src/ndv/views/_qt/_array_view.py
+++ b/src/ndv/views/_qt/_array_view.py
@@ -93,29 +93,6 @@ QRangeSlider { qproperty-barColor: qlineargradient(
 """
 
 
-class QtPopup(QDialog):
-    """A generic popup window."""
-
-    def __init__(self, parent: QWidget | None = None) -> None:
-        super().__init__(parent)
-        self.setModal(False)  # if False, then clicking anywhere else closes it
-        self.setWindowFlags(Qt.WindowType.Popup | Qt.WindowType.FramelessWindowHint)
-
-        self.frame = QFrame(self)
-        layout = QVBoxLayout(self)
-        layout.addWidget(self.frame)
-        layout.setContentsMargins(0, 0, 0, 0)
-
-    def show_above_mouse(self, *args: Any) -> None:
-        """Show popup dialog above the mouse cursor position."""
-        pos = QCursor().pos()  # mouse position
-        szhint = self.sizeHint()
-        pos -= QPoint(szhint.width() // 2, szhint.height() + 14)
-        self.move(pos)
-        self.resize(self.sizeHint())
-        self.show()
-
-
 class _CmapCombo(QColormapComboBox):
     def __init__(self, parent: QWidget | None = None) -> None:
         super().__init__(parent, allow_user_colormaps=True, add_colormap_text="Add...")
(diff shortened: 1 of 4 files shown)


tlambert03:
thanks @gselzer, this would indeed be great to have.   Couple thoughts at this point:

- I'm not sure a context menu is the correct way to show this here,  it gives the strange appearance that the label is "actionable", and generally has the wrong look & feel.  I think something like the QtPopup that I used to show an FPS selector back in v1 would be more appropriate (and generalizable: since I do need to get around to adding back those play buttons and will likely use that again).  https://github.com/pyapp-kit/ndv/blob/b3932383be06f941f6c08176d985eea2f7799b77/src/ndv/v1/_qt/_dims_slider.py#L64-L84
- I would like to be able to set different percentiles for each end.  (Does the mmstudio gui only have one number?  I vaguely remember that being the case and not liking that... particularly with modern scmos cameras, you often want a different high number than low number)

gselzer:
> * I think something like the QtPopup that I used to show an FPS selector back in v1 would be more appropriate (and generalizable: since I do need to get around to adding back those play buttons and will likely use that again).

> * I would like to be able to set different percentiles for each end.

Good ideas. Have updated in the latest commit:

image

If you're a fan of this now I'll investigate how/whether we can do similar things in wx (where I'm more confident) and jupyter (less confident)

marktsuchida:
> * Does the mmstudio gui only have one number?

Yes, but I agree that separate settings are sometimes desirable.

I thought a bit about whether people would want the option of editing just one number, but then again, it doesn't make theoretical sense to use the same number for the read noise at the bottom and the shot noise at the top (for example). It would only be a heuristic convenience. Given that in many cases you'd only need a non-zero percentile for the upper limit (I think?), it seems fine (and simplest) to just have two fields.

gselzer:
@tlambert03 I've been tinkering with a Jupyter version (aided by Claude) over the past few days, but I think I'd like to put that work on a separate branch, as I really want this functionality for qt right now. Any objections to punting on the jupyter functionality for now?

Still planning to investigate wx in the meantime

tlambert03:
> Any objections to punting on the jupyter functionality for now?

not at all.  Similar to the play button, it’s fine with me if features hit different front ends at different times

gselzer:
@tlambert03 I'm pretty happy with these changes, if you'd like to take another look feel free, but can also just merge.

There's one bug I'm aware of but have not yet found a solution for - the tail edit controls for wx do not seem to be editable with keyboard. You can still edit it with the spin buttons (and that will be enough control for e.g. ignoring outliers) so I'm tempted to just merge and file an issue for this - thoughts?

tlambert03:
functionality is great.  I find myself thinking/wondering about the terminology "tail" and "ignore" in  "ignore upper/lower tail".  I'm ok with that, but it's not strictly always "tail shaped"... some alternatives to consider (but if you prefer tail over all of these thats ok too:)

- "clip" or "exclude" or "cutoff" instead of "ignore"
- "percentile" or "darkest/brightest" instead of "tail"?

so:

`Lower Percentile`, `Clip/Exclude Lower Percentile`, `Exclude/Clip Darkest`

thoughts?

gselzer:
> Exclude/Clip Darkest

Yeah, I like `Exclude Darkest`/`Exclude Brightest`! Will change momentarily.

> we could also have a checkbox that toggles all of it, so you can go back to min/max without changing the percentile values?

Maybe - I was originally thinking tabs here, in case we want to add in controls for `ClimsStdDev`, etc. Could also just disable the text edits if the button is toggled off...

tlambert03:
> Maybe - I was originally thinking tabs here, in case we want to add in controls for ClimsStdDev, etc. Could also just disable the text edits if the button is toggled off...

sure, let's consider in another PR

gselzer:
Alright, I'll merge this after renaming the labels and create followup issues for (a) the Jupyter implementation and (b) the wx bug.

PR #155, opened 2025-03-16
tests: update superqt to reduce segfaults

tlambert03:
With https://github.com/pyapp-kit/superqt/pull/283 in, I think our segfaults will be mostly fixed, so this PR ~removes the retries and simplifies ci.yml~

eh... looks like there are a couple residual sporadic segfaults.   so I'm going to leave in the retries.  this just repins superqt, and updates the setup-headless-display action

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,41 +108,13 @@ jobs:
           sudo apt-get update -y -qq
           sudo apt install -y libegl1-mesa-dev libgl1-mesa-dri libxcb-xfixes0-dev mesa-vulkan-drivers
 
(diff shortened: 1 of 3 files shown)


tlambert03:
indeed!  This seems to fix almost everything.  there is now one segfault that seems to occur with vispy and pyside on macos 12... which I'm so far unable to reproduce locally.  https://github.com/pyapp-kit/ndv/actions/runs/13885245165/job/38849778086?pr=155

so that's encouraging!

gselzer:
@tlambert03 thanks for looking into this! Will be nice to have tests passing again!

PR #158, opened 2025-03-20
feat: pygfx gamma

### LINKED ISSUE #78: Enable gamma correction with PyGFX canvas
PyGFX provides a shader-side gamma within the [renderer](https://github.com/pygfx/pygfx/blob/b865a48ae6268bd1a7c4e6e87eb23d25be2c2d61/pygfx/renderers/wgpu/engine/renderer.py#L110, but ideally we'd be able to apply gamma separately to each dataset being displayed (i.e. to each `Image` object). Unless I missed something, I don't think this is currently possible.

Naturally, we can also do it on the CPU, as @tlambert03 suggested in #76, but ideally it'd be on the GPU.

@almarklein do you have any insight here?

gselzer:
Adds gamma support to the pygfx backend:

https://github.com/user-attachments/assets/98ae0446-0a9d-47e7-a9dd-94c3ed892fc3

Closes #78

TODO:
* [ ] Consider gatekeeping PyGFX gamma behind runtime checks instead of bumping the minimum supported version - Thoughts @tlambert03? We've previously gatekept changes ([example](https://github.com/pyapp-kit/ndv/blob/d11d9a0fb4bc0ef5b02ce44e168a9367312fbd3e/src/ndv/views/_pygfx/_array_canvas.py#L384), but this change would take more work to gatekeep...
* [ ] Test

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,7 +76,7 @@ wxpython = [
 
 # Supported Canavs backends
 vispy = ["vispy>=0.14.3", "pyopengl >=3.1"]
-pygfx = ["pygfx>=0.6.0"]
+pygfx = ["pygfx>=0.8.0"]
 
 # ready to go bundles with pygfx
 qt = ["ndv[pygfx,pyqt]", "imageio[tifffile] >=2.20"]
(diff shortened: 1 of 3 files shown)


gselzer:
Also probably best to merge #146 first - much easier to merge those changes into here than these changes into there 😅

tlambert03:
> Consider gatekeeping PyGFX gamma behind runtime checks instead of bumping the minimum supported version

eh... bump the min version.  I'm not so worried about broad backwards pygfx compatibility at the moment.

gselzer:
@tlambert03 happy with this addition? I think it's ready for merge

PR #163, opened 2025-03-26
feat: add back play button, as well as dimension size label

tlambert03:
https://github.com/user-attachments/assets/1cda23ce-d4b1-4599-ad9a-78820fed92ea

### DIFF
--- a/src/ndv/models/_viewer_model.py
+++ b/src/ndv/models/_viewer_model.py
@@ -36,6 +36,8 @@ class ArrayViewerModel(NDVModel):
         Whether to show the ROI button, by default True.
     show_channel_mode_selector : bool, optional
         Whether to show the channel mode selector, by default True.
+    show_play_button : bool, optional
+        Whether to show the play button, by default True.
     show_progress_spinner : bool, optional
         Whether to show the progress spinner, by default
     """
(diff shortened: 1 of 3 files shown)


tlambert03:
@gselzer ... happy enough with this addition?  (still needs to be added to other backends... but mostly I want it in Qt for now)

gselzer:
Oh, another question for you - do you think it makes sense to make the play button a part of the viewer state (i.e. should we add a `ArrayViewerModel.show_play_button`?

tlambert03:
I do!  And actually, even the “play” (Boolean) state and the fps might as well go on the model too?

gselzer:
> And actually, even the “play” (Boolean) state and the fps might as well go on the model too?

Maybe, although I'd think that those would be better suited for the `ArrayDisplayModel` (man, we might want to think about easier names to differentiate)

tlambert03:
ok i added the visibility to the viewer model.  didn't do the play status yet.  can be done later when we decide where to put it.  ready for final review, and you can merge if you're happy with it @gselzer

gselzer:
Looks good to me! Added a quick test for the play button visibility, and filed a few issues for the TODO items.

LGTM

tlambert03:
Thanks!

PR #165, opened 2025-03-30
feat: add XYZ scale to ArrayDisplayModel

tlambert03:
something we've needed here at the QI course...

exposes the ability to set the scale part of the transform (and only the scale for now) on the image being displayed.  mostly needed for anisotropic z sampling


tlambert03:
one challenge this PR generally exposes:  how does a user pass data-related metadata (like channel names, scales, etc...).  Yes, we support named arrays and we try to get all this info from xarray etc...  but i don't want someone to have to create an xarray just to add some metadata.  The `DataWrapper` is the natural place for this to live... but someone shouldn't have to subclass it to customize it (if it's not in the array info itself already).   

this PR could just make the visualization possible, without weighing in on the public API

PR #176, opened 2025-04-16
Histogram hscroll pan

### LINKED ISSUE #172: feature: horizontal scrolling via mouse wheel on histogram
I love that we have proper bounding on the histogram now with #146.  one thing I find myself wanting now is the ability to pan left and right without clicking and draging.  scrolling up and down zooms, I wonder whether scrolling left and right could be used to pan (without making zooming annoying)

gselzer:
This PR enables panning histograms using horizontal scroll in both vispy and pygfx.

TODO:
* [x] Clean up
* [x] Test

Closes #172

### DIFF
--- a/src/ndv/views/_pygfx/_histogram.py
+++ b/src/ndv/views/_pygfx/_histogram.py
@@ -69,6 +69,22 @@ class _OrthographicCamera(pygfx.OrthographicCamera):
         super().set_state(state)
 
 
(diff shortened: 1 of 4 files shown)


gselzer:
@tlambert03 I'm pretty happy with this, works nicely with both mouse and trackpad. Since you requested the feature, thoughts?

PR #179, opened 2025-04-22
feat: Add autoscale tails to Jupyter

gselzer:
Needs cleanup I think, also seems broken ⚠️


gselzer:
@tlambert03 I think this is working well now - do you have time to take another look?

Unfortunately, when you run `examples/notebook.ipynb` autoscale is off by default due to improper signal blocking. However, this also happens on main, so I don't think that bug should hold up a merge here. I *did* put a fair amount of time trying to debug the issue without effort, so I think an issue for later is enough.

PR #181, opened 2025-04-23
feat: histogram highlights

gselzer:
This PR adds a small highlight to indicate the intensity of the image at the cursor position against the histogram:

https://github.com/user-attachments/assets/60017353-660e-4391-bb4a-6413f36a3534

TODO:

* [x] Add support for pygfx
* [x] Test
* [x] Clean
* [x] Make more robust (e.g. sometimes the mouse event can be missed and the highlight stays up even when focus lost)


gselzer:
@tlambert03 I'm pretty happy with this, the codecov failures are due to some event handling code that I don't really care to test. Let me know what you think of the behavior/architecture!

tlambert03 (commented):
functionality and code looks good.

On playing with it a bit, I have to admit, I actually find it a bit distracting 😬 and I'm not sure it adds much to my quantitative understanding of the image ... and so, at the very least, I think I'd want this gated behind an option on the viewer_model (and personally... I'm afraid I'd vote for it to be off by default, but happy to take in some other opinions from @marktsuchida and/or @fdrgsp and be outvoted there)

marktsuchida:
> On playing with it a bit, I have to admit, I actually find it a bit distracting 😬 and I'm not sure it adds much to my quantitative understanding of the image ... and so, at the very least, I think I'd want this gated behind an option on the viewer_model (and personally... I'm afraid I'd vote for it to be off by default, but happy to take in some other opinions from @marktsuchida and/or @fdrgsp and be outvoted there)

I like making it optional, though my preference would be for it to be on by default. There are two things I like about this feature, especially for less experienced users:

- It immediately clarifies some confusions that can easily occur (to beginners, at least) with histograms where there are one or two prominent peaks (containing mostly background), but most of the interesting pixels are in the nearly-invisible tail(s) of the histogram. Say, a single-molecule or in-vitro assay type of fluorescent image. It also helps with odd cases where the field has a gradient and the bright parts of the background is brighter than the dark parts of the objects of interest -- something you'd want to recognize while at the scope (or while _building_ the scope) and address if possible.

- As a teaching tool, it's an easy way to demonstrate how the pixels of the image fall on the histogram. I increasingly notice that not all students immediately grasp the relationship (especially amidst the information overload of learning about a new instrument; it's also harder with some types of images than others).

But I'm sure I'm biased, because I'm the one who added these to MMStudio's histograms.

The reason for preferring on-by-default is that this is something that I think is more useful to newcomers than to experts, and that utility may be lost by the time people discover that this is an option. I'd also be happy with it being off by default in ndv, but on by default in pymmcore-gui, because I think this is more useful during an acquisition session than during offline analysis.

I could imagine some slight adjustment to the brightness or thickness of the line might help with decreasing the distractiveness.

tlambert03:
fair enough!  Yeah, if we could make it slightly less prominent, that would be good.

PR #184, opened 2025-04-25
feat: remove array data display

gselzer:
This PR removes `_ArrayDataDisplayModel`, in favor of storing all contained behavior within the `DataWrapper` and the `ArrayDisplayModel`. More particularly, the `ArrayDisplayModel` is now in charge of slicing the wrapper.

There may be other ways to get rid of the `_ArrayDataDisplayModel` that have merit, and I'm not opposed to experimentation.


src/ndv/models/_data_wrapper.py
```diff
@@ -44,13 +55,39 @@ def shape(self) -> tuple[int, ...]: ...
...
+class DataResponse:
+    """Response object for data requests.
+
+    In the response, the data is broken up according to channel keys.
+    """
+
+    # mapping of channel_key -> data
+    n_visible_axes: int
+    data: Mapping[ChannelKey, np.ndarray] = field(repr=False)
+    request: DataRequest | None = None
```
tlambert03, line 88: these belong in the module that does the slicing, if you're putting the slicing on the array_display_model, then they belong in that module, not here

tlambert03 (commented):
i'm mostly fine with this change.  It does go back to making the Array Display model "heavy" ... which is kinda what I was trying to avoid by adding that additional object: just a thing to negotiate the relationship between the array display model and the data wrapper.   I see this as a relatively parallel adjustment, but I'm ok with it, and that other object has always caused you a good degree of consternation :joy:

Ultimately, I'd like to get the slicing logic back *off* the array display model.  So, this would likely change again in the future. 

two comment that i would want adressed before merge though

gselzer:
I want to think a bit more about this - converting it back to a draft

Issue #191, opened 2025-05-08
feat request: Disable ROI button


gnodar01:
I would like to disable the use of ROI (by removing the roi button).

I'm not entirely sure how the generation of rectangular ROIs is currently used in other contexts, but that is certainly due to my ignorance.

In the context of which we're using NDV, presenting the ability to create ROIs to the user implies to them that there is some dynamic action we can/will take after the ROI is created. Since we take no action, it risks confusing the user.

Again, in the specific context in which we're using NDV, I can certainly see a use in which we generate ROIs programmatically to display to the user. We care more about segmentation masks/labels than rectangular regions, but I could imagine some cases where we would want bounding boxes. However in that case, I wouldn't want the ROIs to be editable by the user (moving/resizing the bounding boxes).

I currently disable  the ROI button (actually just hide, since the underlying signal and observer are still active under the hood) in a hacky way (wx frontend):

```
for child in ndv_viewer.widget().Children:
    if child.GetLabel() == 'ROI':
        child.Destroy()
        break
```

It would be nice to have an api which at removes the button from the frontend widget and/or the ability to toggle/disable it altogether:

```
ndv_viewer = ArrayViewer(..., disable_roi=True)
```

or:

```
ndv_viewer.disable_roi()
```

gselzer:
This feature is [already present](https://github.com/pyapp-kit/ndv/blob/4cb0ec22bcc531e2e5a7a060c032c4cbcb979823/tests/views/_wx/test_array_view.py#L42-L44) in the [`ViewerModel`](https://github.com/pyapp-kit/ndv/blob/4cb0ec22bcc531e2e5a7a060c032c4cbcb979823/src/ndv/models/_viewer_model.py#L37-L38), but as @tlambert03 [said](https://github.com/pyapp-kit/ndv/issues/189#issuecomment-2863130350) over in #189, we could do a better job of exposing it.

gnodar01:
Ah, missed that one. Thanks for pointing it out to me. Resolved on my end, but I'll let you close in case you think its worth keeping around until there's a public api for it. Thanks!

PR #194, opened 2025-05-09
Expose viewer model as `viewer_options` argument to imshow and ArrayViewer

### LINKED ISSUE #189: [feat request] Enable setting cmap value choices
We are able to set the initial value for the `cmap` via `default_lut` and `luts` kwargs in `ArrayWrapper`.

However I would like to be able to specify the full range of options in the dropdown.

For instance, in the wx frontend, this would mean being able to specify the list in the `_WxLUTWidget.cmap` combo box.

Right now I achieve this in a very hacky way that is sure to break in the future:

```
def _set_channels_hack():
    lut_dict = ndv_viewer._view._luts
    for ch_idx in lut_dict.keys():
        # lut_choices = ['magma', 'cyan']
        wx_combo = lut_dict[ch_idx]._wxwidget.cmap.Append(lut_choices)

call_later(1000, _set_channels_hack)
```

I'm sure using `call_later` with a hardcoded timeout will bite me even if the use of non-public attributes doesn't.

tlambert03:
closes #189 and #191
 and probably other issues.

This exposes the `ArrayViewerModel` we've been jealously guarding :joy: ... as an option `viewer_options` to imshow and to `ArrayViewer`.

### DIFF
--- a/src/ndv/util.py
+++ b/src/ndv/util.py
@@ -12,21 +12,32 @@ if TYPE_CHECKING:
 
     from .models._array_display_model import ArrayDisplayModel, ArrayDisplayModelKwargs
     from .models._data_wrapper import DataWrapper
+    from .models._viewer_model import ArrayViewerModel, ArrayViewerModelKwargs
 
 
 @overload
 def imshow(
-    data: Any | DataWrapper, /, display_model: ArrayDisplayModel = ...
+    data: Any | DataWrapper,
+    /,
+    *,
+    viewer_options: ArrayViewerModel | ArrayViewerModelKwargs | None = ...,
+    display_model: ArrayDisplayModel = ...,
 ) -> ArrayViewer: ...
 @overload
 def imshow(
-    data: Any | DataWrapper, /, **kwargs: Unpack[ArrayDisplayModelKwargs]
+    data: Any | DataWrapper,
+    /,
+    *,
+    viewer_options: ArrayViewerModel | ArrayViewerModelKwargs | None = ...,
+    **display_kwargs: Unpack[ArrayDisplayModelKwargs],
 ) -> ArrayViewer: ...
 def imshow(
     data: Any | DataWrapper,
     /,
+    *,
+    viewer_options: ArrayViewerModel | ArrayViewerModelKwargs | None = None,
     display_model: ArrayDisplayModel | None = None,
-    **kwargs: Unpack[ArrayDisplayModelKwargs],
+    **display_kwargs: Unpack[ArrayDisplayModelKwargs],
 ) -> ArrayViewer:
     """Display an array or DataWrapper in a new `ArrayViewer` window.
 
@@ -39,16 +50,26 @@ def imshow(
         The data to be displayed. Any ArrayLike object or an `ndv.DataWrapper`.
     display_model: ArrayDisplayModel, optional
         The display model to use. If not provided, a new one will be created.
-    kwargs : Unpack[ArrayDisplayModelKwargs]
+    viewer_options: ArrayViewerModel | ArrayViewerModelKwargs, optional
+        Either a [`ArrayViewerModel`][ndv.models.ArrayViewerModel] or a dictionary of
+        keyword arguments used to create one.
+        See docs for [`ArrayViewerModel`][ndv.models.ArrayViewerModel] for options.
+    **display_kwargs : Unpack[ArrayDisplayModelKwargs]
         Additional keyword arguments used to create the
-        [`ArrayDisplayModel`][ndv.models.ArrayDisplayModel].
+        [`ArrayDisplayModel`][ndv.models.ArrayDisplayModel]. (Generally, this is
+        used instead of passing a `display_model` directly.)
 
     Returns
     -------
     ArrayViewer
         The `ArrayViewer` instance.
     """
-    viewer = ArrayViewer(data, display_model, **kwargs)
+    viewer = ArrayViewer(
+        data,
+        display_model=display_model,
+        viewer_options=viewer_options,
+        **display_kwargs,
+    )
     viewer.show()
 
     run_app()
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,7 +10,7 @@ repos:
       - id: validate-pyproject
 
   - repo: https://github.com/crate-ci/typos
-    rev: v1
+    rev: v1.32.0
     hooks:
       - id: typos
         args: [--force-exclude] # omitting --write-changes
--- a/examples/numpy_arr.py
+++ b/examples/numpy_arr.py
@@ -14,4 +14,4 @@ except Exception as e:
     print(e)
     img = ndv.data.nd_sine_wave((10, 3, 8, 512, 512))
 
-viewer = ndv.imshow(img, viewer_options={"default_luts": ["gray", "green", "magenta"]})
+viewer = ndv.imshow(img)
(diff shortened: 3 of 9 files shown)


src/ndv/util.py
```diff
@@ -12,21 +12,32 @@
...
 def imshow(
     data: Any | DataWrapper,
     /,
+    *,
+    viewer_options: ArrayViewerModel | ArrayViewerModelKwargs | None = None,
     display_model: ArrayDisplayModel | None = None,
-    **kwargs: Unpack[ArrayDisplayModelKwargs],
+    **display_kwargs: Unpack[ArrayDisplayModelKwargs],
 ) -> ArrayViewer:
     """Display an array or DataWrapper in a new `ArrayViewer` window.
```
gselzer, line 42: It'd be cool to enable something like:
```python
ndv.imshow(img, visible_axes=(2, 3), show_roi_button=False)
```
What about replacing all of these with just
```python
def imshow(
    data: Any | DataWrapper,
    /,
    *,
    viewer_model: ArrayViewerModel | ArrayViewerModelKwargs | None = None,
    display_model: ArrayDisplayModel | ArrayDisplayModelKwargs | None = None,
    **kwargs: Unpack[ModelKwargs],
) -> ArrayViewer:
```

where

```
class ModelKwargs(ArrayViewerModelKwargs, ArrayDisplayModelKwargs):
    pass
```

I suppose maybe with this maybe you could pass an `ArrayDisplayModel` as well as a `visible_axes`?, but that can happen already with the bottom signature. 

I did spend some time goofing around with smarter options (i.e. you can just pass a ViewerModel and a DisplayModel or a DisplayModel and some Viewer model args or a ViewerModel and some Display model args or just a bunch of viewer/display model args) but wasn't successful there...

gselzer (commented):
Thanks for doing this @tlambert03! Couple small comments.

Also, now that we are exposing the `ArrayViewerModel`...do you think that we differentiate our names enough? i.e. is `ArrayViewerModel` different enough from `ArrayDisplayModel` (and also from `ArrayViewer`?) Seeing as how the `ArrayViewerModel` has little to do with the array itself...I wonder if just `ViewerModel` would be better?

tlambert03:
Again, I'm sort of reserving the bare "ViewerX" for a new widget that deals with more than a single array.  So, I'd rather be able to just say "this complicated name is now replaced with this more simple name" than have to come up with  another plain "Viewer" name in the future.  Inasmuch as we're dealing with a single array, no transforms, no overlays... we just preface everything with `Array`


As for "viewer" vs "display" that could indeed take some more thought.  Not sure

tlambert03:
Again, I'm sort of reserving the bare "ViewerX" for a new widget that deals with more than a single array.  So, I'd rather be able to just say "this complicated name is now replaced with this more simple name" than have to come up with  another plain "Viewer" name in the future.  Inasmuch as we're dealing with a single array, no transforms, no overlays... we just preface everything with `Array`


As for "viewer" vs "display" that could indeed take some more thought.  Not sure

gselzer:
> Again, I'm sort of reserving the bare "ViewerX" for a new widget that deals with more than a single array.

Well, the point I was making was that `ArrayViewerModel`  as it is has nothing (very little? e.g. `show_roi_button` may be a bit limiting) limiting it to a single array. If you want to wait, that's fine, however I think it's problematic that as a developer of these models I often have trouble remembering which is which, and this PR now makes this problem public.

> As for "viewer" vs "display" that could indeed take some more thought. Not sure

`(Array)UIModel` may be one option.

tlambert03:
> (Array)UIModel may be one option.

would that alleviate the confusion and make it easier to remember?  That’s fine with me.

> as a developer of these models I often have trouble remembering which is which

Internally, I say to myself “the array display model tells me how the array should be displayed, and the array viewer model tells me how the viewer should be configured”.  I guess the bit in brackets there is that the viewer is a UI component?  Is that the bit that’s hard?

gselzer:
> Internally, I say to myself “the array display model tells me how the array should be displayed, and the array viewer model tells me how the viewer should be configured”. I guess the bit in brackets there is that the viewer is a UI component? Is that the bit that’s hard?

"Display" and "Viewer" are synonyms, so the bit that is hard is having to rely upon our existing terminology conventions to remember which does which. Ideally, the model names would not allow for this confusion.

Can we switch `ArrayViewerModel` to `ArrayUIModel`? Open to other names too...

tlambert03:
> "Display" and "Viewer" are synonyms

I dunno... I would say display and view are synonyms more than display and viewER.  I guess that's why it's all quite fine for me.  a viewer displays (it isn't in itself *the* display ... it's the thing whose coordinated action _generates_ a display)

this really does feel like some terrible bike shedding though.  The primary point here, and the one that I'm not wanting to compromise on just yet, is what seems to be a desire to use names that will persist "indefinitely", even if we move to a non-single-array type of widgets.  That's the thing I really *do* want to wait on, because, as mentioned earlier, it's far far easier to say "ok, you can just change the name to plain old `Thing` now, without changing your API" ... and so, your observation that there's really nothing limiting this to a single array is not particularly compelling to me: because we're just not there yet; and there's no compelling reason to force a more permanent name.  We can very easily switch that under the hood in the future, and just give a warning about a deprecated name (you can't, go the other way, however, without a breaking change)

src/ndv/util.py
gselzer: > yeah, could be done. however, the reason for the overload was more to indicate that you shouldn't pass both kwargs and a model (I don't want to get into the business of magical merging of the two and deciding which takes priority and stuff).

Yup, I figured that was what you were trying to prevent, and I agree it would be good to avoid reconciling a model and kwargs. The problem though is that the current changeset is already open to that; due to that final signature, you can write `ndv.imshow(data, display_model=my_display_model, channel_axis=3)` and typing wise that's okay.

tlambert03: > The problem though is that the current changeset is already open to that; due to that final signature, you can write ndv.imshow(data, display_model=my_display_model, channel_axis=3) and typing wise that's okay.

but how is otherwise not ok?  i.e. what's the problem?

tlambert03: ohh... sorry I thought you were talking about viewer_options,  are you saying that it's typing wise ok to use display_model and channel_axis at the same time?

(it' not):

Screenshot 2025-05-15 at 10 19 04 AM

tlambert03:
> Can we switch ArrayViewerModel to ArrayUIModel? Open to other names too...

i guess my main argument against that is the question "What is the `ArrayUI`"?  It's pretty clear what `ArrayViewer` is: it's the primary object: `ndv.ArrayViewer` and `ArrayViewerModel` is the model that describes its state.  `ArrayUI` ... is that the user interface for the array itself?

gselzer:
> this really does feel like some terrible bike shedding though.

I think a little bikeshedding is in order whenever we add new public API, but yes, maybe I'm being too aggressive 😅

> i guess my main argument against that is the question "What is the `ArrayUI`"? It's pretty clear what `ArrayViewer` is: it's the primary object: `ndv.ArrayViewer` and `ArrayViewerModel` is the model that describes its state. `ArrayUI` ... is that the user interface for the array itself?

My goal is to avoid confusion about the different models exposed by ndv, and my argument is that `ArrayDisplayModel` and `ArrayViewerModel` are similar enough to cause that confusion. My issue is that I can't come up with anything better, and this (now that I have some caffeine in my system): 

> > Can we switch ArrayViewerModel to ArrayUIModel? Open to other names too...
> 
> i guess my main argument against that is the question "What is the `ArrayUI`"? It's pretty clear what `ArrayViewer` is: it's the primary object: `ndv.ArrayViewer` and `ArrayViewerModel` is the model that describes its state. `ArrayUI` ... is that the user interface for the array itself?

makes perfect sense to me. So let's just move forward with this solution as is until either (a) someone else complains or (b) I get some better ideas. Does that sound good to you?

PR #195, opened 2025-05-13
Better gui for many channels (wx-specific)

### LINKED ISSUE #193: [feat request] Implement a way to organize composite view on many-channel images
When viewing a many-channel image (e.g. [this one](https://www.ebi.ac.uk/biostudies/bioimages/studies/S-BIAD1236)) the composite mode view gets quite cluttered.

Image

The simplest solution would be to have the `_WxArrayViewer` use a scrollable area with a parameterized max height.

What would be nice to have is a dropdown with multiselect (or some similar UI) allowing me to choose e.g. 2-5 channels for compositing, and having the controls for the rest hidden.

gnodar01:
- Creates a scrollable area for LUTs
- Creates a multi-select to choose which LUTs go in the scrollable area

Resolves #193

### DIFF
--- a/src/ndv/views/_wx/_array_view.py
+++ b/src/ndv/views/_wx/_array_view.py
@@ -67,6 +67,142 @@ def _add_icon(btn: wx.AnyButton, icon: str) -> None:
     btn.SetLabel("")
     btn.SetBitmapLabel(bitmap)
 
+class _LutChannelSelector(wx.Panel):
+    def __init__(self, parent: wx.Window, channels: None | list[ChannelKey]=None):
+        super().__init__(parent)
+
+        self.channels = channels or []
+        # all channels visible, by default
+        self.visible_channels = set(self.channels)
+
+        # Dropdown button with currenct selection
+        self.dropdown_btn = wx.Button(
+            self, label="Channel Display Options", style=wx.BU_EXACTFIT)
+        self.dropdown_btn.Bind(wx.EVT_BUTTON, self._on_dropdown_clicked)
+
+        # Display indicator for how many channels are currently visible
+        self.selection_info = wx.StaticText(self, label="")
+        self._update_selection_info()
+
+        # Create a popup window for the checklist
+        self.popup = wx.Dialog(self, style=wx.BORDER_SIMPLE | wx.STAY_ON_TOP | wx.FRAME_NO_TASKBAR | wx.FRAME_FLOAT_ON_PARENT)
+        self.popup.SetBackgroundColour(wx.SystemSettings.GetColour(wx.SYS_COLOUR_WINDOW))
+        # evt handler for clicking outside of popup dialog
+        self.popup.Bind(wx.EVT_ACTIVATE, self._on_popup_deactivate)
+
+        # Create a checklist for the popup
+        self.checklist = wx.CheckListBox(self.popup, choices=[])
+        self.checklist.Bind(wx.EVT_CHECKLISTBOX, self._on_selection_changed)
+
+        # Select All / None buttons
+        btn_sizer = wx.BoxSizer(wx.HORIZONTAL)
+        self.select_all_btn = wx.Button(self.popup, label="Select All", size=(80, -1))
+        self.select_none_btn = wx.Button(self.popup, label="Select None", size=(80, -1))
+        self.select_all_btn.Bind(wx.EVT_BUTTON, self._on_select_all)
+        self.select_none_btn.Bind(wx.EVT_BUTTON, self._on_select_none)
+        btn_sizer.Add(self.select_all_btn, 0, wx.ALL, 5)
+        btn_sizer.Add(self.select_none_btn, 0, wx.ALL, 5)
+
+        # Layout for popup
+        popup_sizer = wx.BoxSizer(wx.VERTICAL)
+        popup_sizer.Add(self.checklist, 1, wx.EXPAND | wx.ALL, 5)
+        popup_sizer.Add(btn_sizer, 0, wx.ALIGN_CENTER | wx.BOTTOM, 5)
+        self.popup.SetSizer(popup_sizer)
+
+        # Layout for the main widget
+        sizer = wx.BoxSizer(wx.HORIZONTAL)
+        sizer.Add(self.dropdown_btn, 0, wx.ALIGN_CENTER_VERTICAL | wx.RIGHT, 5)
+        sizer.Add(self.selection_info, 0, wx.ALIGN_CENTER_VERTICAL)
+        self.SetSizer(sizer)
+
+        # Signals
+        self.on_selection_changed_callback = None
+
+    def set_channels(self, channels: list[ChannelKey]):
+        """Update the channel list."""
+        self.channels = list(channels)
+        self.visible_channels = set(self.channels)
+        self._update_checklist()
+        self._update_selection_info()
+
+    def _update_checklist(self):
+        """Update the checklist with current channels."""
+        self.checklist.Clear()
+        for channel in self.channels:
+            self.checklist.Append(str(channel))
+            idx = self.checklist.FindString(str(channel))
+            if channel in self.visible_channels:
+                self.checklist.Check(idx)
+
+    def _update_selection_info(self):
+        """Update the selection info text."""
+        if len(self.visible_channels) == len(self.channels):
+            self.selection_info.SetLabel(
+                f"All channels visible ({len(self.channels)})")
+        else:
+            self.selection_info.SetLabel(
+                f"{len(self.visible_channels)} of {len(self.channels)} channels visible")
+        self.Layout()
+
+    def _on_dropdown_clicked(self, evt):
+        """Show the dropdown checklist."""
+        # Position the popup below the button
+        btn_size = self.dropdown_btn.GetSize()
+        btn_pos = self.dropdown_btn.GetPosition()
+        popup_pos = self.ClientToScreen(
+            wx.Point(btn_pos.x, btn_pos.y + btn_size.height))
+
+        # Update the checklist
+        self._update_checklist()
+
+        # Size the popup according to content
+        self.popup.SetSize(wx.Size(250, min(400, 50 + len(self.channels) * 22)))
+        self.checklist.SetSize(self.popup.GetClientSize())
+        self.popup.Layout()
+
+        # Show the popup
+        self.popup.SetPosition(popup_pos)
+        self.popup.Show(show=True)
+
+    def _on_popup_deactivate(self, evt):
+        """Close the popup when it loses focus (user clicks outside)."""        
+        if not evt.GetActive():
+            self.popup.Show(show=False)
+
+    def _on_selection_changed(self, evt):
+        """Handle selection changes in the checklist."""
+        idx = evt.GetSelection()
+        channel = self.channels[idx]
+
+        if self.checklist.IsChecked(idx):
+            self.visible_channels.add(channel)
+        else:
+            self.visible_channels.discard(channel)
+
+        self._update_selection_info()
+
+        # Notify parent
+        if self.on_selection_changed_callback:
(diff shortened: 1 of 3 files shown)


gnodar01:
@tlambert03 @gselzer Implemented what was requested in #193.

Uploading Screen Recording 2025-05-13 at 5.57.49 PM.mov…


Draft because it could use a bit of cleanup, but also there's a bit of a problem around lifecycle control. In the video you can see when I switch back to composite mode, all of the channels are visible instead of only the ones selected in the options. 

First channel mode is `WxArrayView.set_channel_mode` is called and then after each `WxLutView.set_visible` is called. If this happened in the opposite order I could emit an event after `set_channel_mode` to call `set_visible` on only the luts that are selected in the multiselect options (at the cost of having two calls of `set_visible` per LUT view per mode change).

There's other approaches, but I wanted to avoid a state injection on `WxLutView` that conditions `set_visible` on whether its enabled in the multiselect options.

Plenty of other approaches, but wanted to get your thoughts first. TBH the scroll area gets me most of what I want anyway, so I wouldn't bemoan getting rid of the multiselect in general. But if you like it, any advice on how we could change up the lifecycle, or whether that's the wrong way to think about it?


Also haven't looked into putting this into the tests yet, but any pointers would be appreciated.

gselzer:
> Also haven't looked into putting this into the tests yet, but any pointers would be appreciated.

Oh, and by the way, I think the place to test this behavior would be [here](https://github.com/pyapp-kit/ndv/blob/main/tests/views/_wx/test_array_view.py)

gnodar01:
Of course! Thanks for saving me a ton of time by creating ndv.

Okay, so I ended up taking a different approach such that `WxLutView.set_visible` now emits rather than calling `Show/Hide` on the widget directly. Then we validate later in `WxArrayView` and call `Show/Hide` only if necessary.

When reviewing, please do pay attention to my use of `Signal`. It's a really great tool (love the observer pattern), but I fear I may have abused it.

Will add tests in the meantime.

src/ndv/views/_wx/_array_view.py
```diff
@@ -308,14 +469,17 @@ def set_clim_bounds(
...
-            self._wxwidget.Show()
-        else:
-            self._wxwidget.Hide()
+    def set_visible(self, visible: bool, validate: bool = True) -> None:
+        """Sets visibility.
+
+        `validate` will check the display options (`_LUTChannelSelctor`)
+        to ensure that visibility is toggled off there.
+        """
+        self.lutUpdated.emit(self._channel, visible, validate)
```
tlambert03, line 478: yeah, I do find this pattern a little backwards.  signal emitters shouldn't really "depend" on their callbacks at all.  So this method, that appears to be setting the visibility of a given `WxLutView`, really isn't doing that at all unless some other object is "out there listening" to these events and then mutating this object on its behalf.

In this case, it's `WxArrayView` that is invoking this `set_visible` method (inside of `WxArrayView._on_lut_selection_changed`,  *and* it's `WxArrayView` that is also the sole receiver of the signal emission via its callback (`WxArrayView._update_lut_view`), which takes care of actually setting the visibility of that LutView by calling `self._luts[channel]._wxwidget.Show/Hide()` ... So why does this method and signal need to exist at all?  Why not just have `WxLutView._on_lut_selection_changed` do it directly?

tlambert03 (commented):
just a few high level observations and thoughts, and a couple inline nits

- if there are only a few channels,  that "channel display options" row takes up some unfortunate space.  So I wonder whether the whole behavior (or at least the visibility of that option) should be gated behind some minimum number of channels?
- i wonder whether toggling the *inclusion* of a channel in the channel display options should also toggle the visibility of the channel itself?  (that is, do we ever want a channel to be visible on the canvas that doesn't show up in the list of channels)?  I noticed that by trying it with a 16 channel image, choosing "select none", and being a little confused at first at what I was looking at, and why it had so many channels 
- not for this PR, but a note for my future self. the relatively slow performance with 16 channels, *even* when only one of them is showing at all, makes me wonder whether we're making unnecessary slicing and/or rendering calls for non-visible textures

gnodar01:
> if there are only a few channels, that "channel display options" row takes up some unfortunate space. So I wonder whether the whole behavior (or at least the visibility of that option) should be gated behind some minimum number of channels?

Good idea, will add that.


> i wonder whether toggling the inclusion of a channel in the channel display options should also toggle the visibility of the channel itself? (that is, do we ever want a channel to be visible on the canvas that doesn't show up in the list of channels)? I noticed that by trying it with a 16 channel image, choosing "select none", and being a little confused at first at what I was looking at, and why it had so many channels

Didn't even think about that. Will add that as well.

src/ndv/views/_wx/_array_view.py
gnodar01: Totally agree. I took a problematic approach. I'll revisit it. Please feel free to totally ignore the wall of text below, I'm just a fan of documenting my thought process because my memory is horrible.

---

The reasoning was:

I wanted to make changes that were totally self-contained in the `_wx/_array_view.py`, so that none of the code elsewhere in `ndv` had to change, particularly the controller. From the controllers point of view, all views should have the same API.

That's when I ran into the lifecycle problem [above](https://github.com/pyapp-kit/ndv/pull/195#issuecomment-2878068338).

The problem was that we would hide some channels in the display options, triggering:
- `_LutChannelSelector._on_selection_changed` -> `emit`
- `observe` -> `WxArrayView._on_lut_selection_changed`
- `WxLutView.set_visible`

Then we'd change mode from `COMPOSITE` to `GRAYSCALE`. This would cause the controller to call `WxLutView.set_visible(False)` on each LUT (via `controllers/_array_viewer::ArrayViewer._on_model_channel_mode_changed`).

Then we'd change mode from `GRAYSCALE` to `COMPOSITE`. This would cause the controller to call `WxLutView.set_visible(True)` on each LUT.

Which would mean all the LUTs are now visible, even if they are deselected in the display options.

So I needed some way to validate whether they were disabled in the LUT channel selector first, before allowing the controller call `Show/Hide`.

But the individual `WxLutView` objects don't have access to the LUT channel selector, and they don't know whether they are enabled/disabled in the LUT channel selector. That's what I meant by this above:

> ...I wanted to avoid a state injection on `WxLutView` that conditions `set_visible` on whether its enabled in the multiselect options.

I wanted to avoid putting a new field in `WxLutView` like `displayed: bool`.

`_WxArrayViewer` does have this information, since it contains the `_LutChannelSelector`. So I needed `_WxArrayViewer` to handle the validation, and then make the final decision to `Show/Hide`.

So that's why I have `WxLutView.set_visible` emit a signal instead of calling `Show/Hide` directly.

As for `WxArrayView` both calling `WxLutView.set_visible` and receiving the signal that causes it to actually call `Hide/Show`, rather than it just calling `self._update_lut_view` directly, that's somewhat of an oversight. But I think my instinct was to go through the "official" process of calling `WxLutView.set_visible`, just as the controller does. That way in the future if `set_visible` does something extra, the behavior between the controller's use of it and the LUT selector's use of it is the same.

gnodar01: UPDATE: refactored ✅

I ended up putting a bit of extra state in `WxLutView` after all via an `Enum` that tracks the three states it can be in: `VISIBLE` (controls shown, visibility on), `DISPLAYED` (controls shown, visibility off), `HIDDEN` (controls not shown, visibility off).

gnodar01:
@tlambert03 @gselzer Ready for review, and merge if no problems. Happy to address any additional changes as well.

src/ndv/views/_wx/_array_view.py
```diff
@@ -566,10 +835,26 @@ def add_histogram(self, channel: ChannelKey, canvas: HistogramCanvas) -> None:
...
+
+        # Remove from our dictionaries
+        if channel_to_remove is not None:
+            del self._luts[channel_to_remove]
+
+        self._lut_selector().remove_channel(channel)
+        self._wxwidget.update_lut_scroll_size()
+
+        if len(self._luts) < self._wxwidget._toolbar_display_thresh:
+            self._wxwidget.toggle_lut_toolbar(False)
```
gselzer, line 857: I'm guessing there's a potential bug here where you have `self._wxwidget._toolbar_display_thresh` channels, some of which are not displayed. If you then remove a lut view to go below the threshold, making the toolbar vanish, you then have no UI controls to visualize the channel?

...I don't think this method is currently called by anything though, so I don't know, maybe just throw in a `FIXME`?

tests/views/_wx/test_array_view.py
```diff
@@ -62,11 +70,133 @@ def processEvent(evt: wx.PyEventBinder, wdg: wx.Control) -> None:
...
+    assert viewer._luts[0]._wxwidget.IsShown()
+
+    # channel_mode = viewer._wxwidget.channel_mode_combo
+    # viewer.set_channel_mode(ChannelMode.GRAYSCALE)
+    # _processEvent(wxapp, wx.EVT_COMBOBOX, channel_mode)
+
+    ## all channels should be hidden
+    # for ch, lut_view in viewer._luts.items():
+    #    if type(ch) is int or (type(ch) is str and ch.isdigit()):
+    #        assert not lut_view._wxwidget.IsShown()
```
gselzer, line 141: Just wondering about the status of this code. Is this something you wanted to keep or move somewhere? Or does it not pass (yet)?

gselzer (commented):
This has made some nice steps forward since I last looked! Thanks again @gnodar01!

> I have it so it automatically sets visibility off when display is off. When display is set to on, you then have to set visibility on manually. That felt more natural to me, but can easily make it so display on triggers visibility on too.

I actually feel the opposite, although I wouldn't say I feel terribly strongly:
* If you click "Select None" and then "Select All", it's unfortunate that to actually see the data you'd then have to go through and click all of the "Visible" checks for each channel. 
* If your goal is just to visualize Channel 5 (of some large number of channels), you might click "Select None", and then you click on Channel 5 in the "Channel Display Options". But then there's the additional inconvenience of checking the visible box.

>  (that is, do we ever want a channel to be visible on the canvas that doesn't show up in the list of channels)?

There's a lot of overlap, isn't there? 😅

In fact, I think another question has merit: Do we ever want a channel to show up in the list of channels if it is not visible on the canvas?

Certainly, if you're below the threshold required to show the LUT selector, then you need all of the channels in the list.

However, whenever the LUT selector is present, the two checkboxes provide a bit of redundancy, and a possible source of bugs, no?

I know it's late in the development process here (especially those tests, very nice 😍), so feel free to disregard,wever I was inspired to draw a potential alternative design, which would only expose a single source of "LUT visibility" - thoughts @tlambert03, @gnodar01?
thumbnail_1000007190

gselzer:
Aww, man, double review?! Sorry! 🫣

tlambert03:
> > Arguably the logic for choosing which channels to composite should be implemented outside of NDV, but thought I'd raise it here for consideration. Would really like a scrollable area though.

> This does resonate more with me now. In light of the complexity we've described in the posts above, is there reason to keep the channel options logic in the UI? i.e. why not slice the data outside of NDV?

I actually don't think it should have to be implemented outside of NDV.  it's very much inline with all of the other things we do where "we make it convenient to view high dimensional data."  You could also say that someone could just slice their XYZT data outside of ndv and give us the final result for viewing, but we're definitely not making that argument.  So I'm all for making high channel dimensions more convenient as well.  This doesn't have to be the final, ultimate UI that we land on, and we can keep reconsidering patterns as we play with it, but I'm for it 👍 

sorry for the delay on this @gnodar01 ... got side-tracked on another project this weekend.

tlambert03:
fixed pre-commit typing errors in 98ac036ff7571585db5aa750e5e5e034ec977046

gnodar01:
Thanks @tlambert03 @gselzer! Didn't have time to come back to this until just now. If you do end up on a different pattern elsewhere, and want help implementing it on the WX frontend, feel free to ping me. Would be happy to help.

PR #196, opened 2025-05-16
feat: Purge ArrayDataDisplayModel from ArrayViews

gselzer:
The goal of this PR is to minimize the footprint of the `ArrayDataDisplayModel` by removing its usage within the `ArrayView`s. It is a potential alternative/complement to #184.

It turns out that the only reason the `ArrayDataDisplayModel` is being passed to the views is so that the array views could decide on the best dimension to visualize in the case we were moving from 2 dimensions to 3. I think this design is suboptimal as:
* It allows business logic in the views, when that logic should really be in the controller.
* That logic must be present in each `ArrayView` (i.e. it's WET)

The reason I hesitate with this change though is because I'd love a world where all views in NDV know about and operate directly on their associated models; this change goes against that by removing models (granted, models I dislike 😅) from the views, instead propagating desires indirectly to the model via signals. However, I think there's a difference between the model-view interactions involved in #87:
* With LUTs right now, the interactions are just "model parameter X should be Y"
* Here, the interaction is "model parameter `visible_axes` should be 3 items long" - information is still missing, which needs to be filled in. In that respect, it's like all of the other operations where we use signals.

So I'm not sure whether it's a final solution, but I do hope it's a step towards one by restricting the usage of `ArrayDataDisplayModel` for later refactoring.

### DIFF
--- a/src/ndv/controllers/_array_viewer.py
+++ b/src/ndv/controllers/_array_viewer.py
@@ -435,7 +435,7 @@ class ArrayViewer:
         self._data_model.display.current_index.update(self._view.current_index())
 
     def _on_view_ndims_requested(self, ndims: int) -> None:
-        """Update the model when the visible axes change."""
+        """Update the model when the user requests ndims visualized axes."""
         # TODO: Disable 3D mode when only 2 axes?
         current_axes = self.display_model.visible_axes
         if len(current_axes) > 2:
--- a/src/ndv/controllers/_array_viewer.py
+++ b/src/ndv/controllers/_array_viewer.py
@@ -101,9 +101,7 @@ class ArrayViewer:
 
         # TODO: Is this necessary?
         self._histograms: dict[ChannelKey, HistogramCanvas] = {}
-        self._view = frontend_cls(
-            self._canvas.frontend_widget(), self._data_model, self._viewer_model
-        )
+        self._view = frontend_cls(self._canvas.frontend_widget(), self._viewer_model)
 
         self._roi_view: RectangularROIHandle | None = None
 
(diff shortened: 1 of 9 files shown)


tlambert03 (approved):
I like this change, let's go with it.  

I'll note that I don't have a problem (at all) with the ArrayDataDisplayModel thing.  I see it not as an evil pattern with a code smell, but just as some private internal object that coordinates a couple things.  Perhaps the word "model" is the thing that bothers you about it?  it really is more of a controller.  But it's a necessary controller that combines things that shouldn't otherwise be together elsewhere.  

In keeping with it as a controller, I like that this PR moves its controller-like functionality off of the views and back to our main controller.  

I suspect that we will later add at least some model (`ArrayDisplayModel`) *back* to the init in the future, and so in that regard, I see the change of the view's `__init__` as a temporary change until the next time we need it again.  but I have no problems removing it until that time arises.

I do think the test failures are legitimate though...  when I run this locally:

```sh
uv run -p 3.12 --exact --no-dev --extra=wxpython --extra=pygfx  '--group=test' pytest
```

the tests pass, and then I get a "python exited unexpectedly" on cleanup.  can you see if you can fix that?

gselzer:
> Perhaps the word "model" is the thing that bothers you about it? it really is more of a controller. But it's a necessary controller that combines things that shouldn't otherwise be together elsewhere.

You know, I think you've hit on why I dislike it! Definitely seems like more of a controller than a model. But yes, agree the functionality is necessary and must live somewhere.

> I do think the test failures are legitimate though... when I run this locally:
> 
> ```shell
> uv run -p 3.12 --exact --no-dev --extra=wxpython --extra=pygfx  '--group=test' pytest
> ```

Interesting, hadn't actually noticed the failure yet...but I'll take a look soon!

gselzer:
@tlambert03 got the tests fixed, good to merge on my end!

PR #202, opened 2025-05-28
fix: update hover on new data

### LINKED ISSUE #199: Histogram highlight does not update when displayed data changes
I'm sure the controller logic just needs a simple update to query the data more often:

Image

gselzer:
Closes #199.

We need to update the hover info whenever:
1. The user moves the mouse
2. The displayed data is updated under the mouse

Only the first case was previously covered.

To cover (2), we must keep track of the mouse position. Currently, this state is stored within the `ArrayViewer` and updated whenever (1) occurs. This state *could* potentially live in the `ArrayViewerModel` instead, if we find it helpful for other tasks.

### DIFF
--- a/src/ndv/controllers/_array_viewer.py
+++ b/src/ndv/controllers/_array_viewer.py
@@ -117,7 +117,7 @@ class ArrayViewer:
         self._view.channelModeChanged.connect(self._on_view_channel_mode_changed)
         self._view.nDimsRequested.connect(self._on_view_ndims_requested)
 
-        self._last_mouse_pos: tuple[int, int] | None = None
+        self._highlight_pos: tuple[int, int] | None = None
         self._canvas.mouseMoved.connect(self._on_canvas_mouse_moved)
         self._canvas.mouseLeft.connect(self._on_canvas_mouse_left)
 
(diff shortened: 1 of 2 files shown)


gselzer:
@tlambert03 can you replicate those widget leaks locally? I was actually able to replicate up until I made some local changes. Now, even after undoing them, I can no longer replicate...

tlambert03:
I've seen those recently popping up occasionally and sporadically.  Note, whenever you see a "widget leak" error, you need to check whether some other error comes before it.  And in this case, the widget leak is just a failed cleanup after the *actual* error that is shown a bit higher (https://github.com/pyapp-kit/ndv/actions/runs/15305150376/job/43055857610?pr=202#step:11:393)

```
  File "/Users/runner/work/ndv/ndv/.venv/lib/python3.12/site-packages/vispy/gloo/texture.py", line 324, in _resize
    shape = self._normalize_shape(shape)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/runner/work/ndv/ndv/.venv/lib/python3.12/site-packages/vispy/gloo/texture.py", line 193, in _normalize_shape
    raise ValueError("Too many channels for texture")
ValueError: Too many channels for texture
ERROR    vispy:logs.py:327 Invoking <bound method SceneCanvas.on_draw of <SceneCanvas (PyQt6) at 0x115ec9b80>> for DrawEvent
```

... however, I don't yet know what's causing it

gselzer:
> And in this case, the widget leak is just a failed cleanup after the _actual_ error that is shown a bit higher (https://github.com/pyapp-kit/ndv/actions/runs/15305150376/job/43055857610?pr=202#step:11:393)

Ah, good catch. Yeah, looks like an issue likely arising from the changes added by #196. I'm guessing it's a race condition between a 3D data request and clearing the canvas (i.e. the `ImageHandle` backed by a vispy image is being used for displaying 3D data when it should have been cleared and replaced by a vispy volume `ImageHandle`?)

tlambert03 (approved):

PR #211, opened 2025-08-20
Make image showing functions thread-safer

ctrueden:
This patch improves ndv's image showing functions to work more seamlessly when called from non-main threads, especially on macOS.

Below is an MCVE script that illustrates the problem before this patch. Without this patch, commenting out the `@ensure_main_thread` decorator causes the program to crash, whereas with this patch, the program works with or without the decorator in place.

Note that the patch itself was written by Claude.ai, not me, so I won't be offended by any harsh criticism. 😆 

```python
import sys
import threading
import time

from qtpy.QtWidgets import QApplication


def create_and_show_random_image() -> "ndv.ArrayViewer":
    from ndv import ArrayViewer
    import numpy
    from ndv.views._app import ensure_main_thread

    @ensure_main_thread  # Comment out this decorator to observe the crash.
    def show(data, **kwargs):
        viewer = ArrayViewer(data)
        viewer.show()
        return viewer

    data = numpy.random.random([300, 200])
    retval = show(data)
    return retval.result() if hasattr(retval, 'result') else retval


def do_stuff_later():
    print("I'll do it in a sec...")
    time.sleep(1)
    print("Time to show some images!")
    for _ in range(3):
        print("Showing an image...")
        viewer = create_and_show_random_image()
        viewers.append(viewer)
        print(f"Viewers -> {viewers}")
    print("All images shown.")


# Create QApplication on main thread.
print("Creating Qt application...")
app = QApplication(sys.argv)

viewers = []

threading.Thread(target=do_stuff_later).start()

print("Starting Qt loop...")
app.exec()

print("That's it! Go home to your pets!")
```

### DIFF
--- a/src/ndv/controllers/_array_viewer.py
+++ b/src/ndv/controllers/_array_viewer.py
@@ -104,7 +104,9 @@ class ArrayViewer:
 
         if self._data_model.data_wrapper is not None:
             # Queue view synchronization to main thread as well since it involves GUI operations
-            sync_future = _app.ndv_app().call_in_main_thread(self._fully_synchronize_view)
+            sync_future = _app.ndv_app().call_in_main_thread(
+                self._fully_synchronize_view
+            )
             sync_future.result()
 
     def _init_gui_components(self) -> None:
--- a/src/ndv/controllers/_array_viewer.py
+++ b/src/ndv/controllers/_array_viewer.py
@@ -103,7 +103,8 @@ class ArrayViewer:
         init_future.result()
 
         if self._data_model.data_wrapper is not None:
-            # Queue view synchronization to main thread as well since it involves GUI operations
+            # Queue view synchronization to main thread as well
+            # since it involves GUI operations
             sync_future = _app.ndv_app().call_in_main_thread(
                 self._fully_synchronize_view
             )
--- a/src/ndv/controllers/_array_viewer.py
+++ b/src/ndv/controllers/_array_viewer.py
@@ -97,6 +97,18 @@ class ArrayViewer:
         # where None is the default channel
         self._lut_controllers: dict[ChannelKey, ChannelController] = {}
 
+        # Thread-safe initialization: queue GUI-sensitive operations to main thread
+        init_future = _app.ndv_app().call_in_main_thread(self._init_gui_components)
+        # Block until GUI components are initialized
+        init_future.result()
+
+        if self._data_model.data_wrapper is not None:
+            # Queue view synchronization to main thread as well since it involves GUI operations
+            sync_future = _app.ndv_app().call_in_main_thread(self._fully_synchronize_view)
+            sync_future.result()
+
+    def _init_gui_components(self) -> None:
+        """Initialize GUI components that must be created on the main thread."""
         # get and create the front-end and canvas classes
         frontend_cls = _app.get_array_view_class()
         canvas_cls = _app.get_array_canvas_class()
@@ -120,9 +132,6 @@ class ArrayViewer:
         self._highlight_pos: tuple[int, int] | None = None
         self._canvas.mouseMoved.connect(self._on_canvas_mouse_moved)
         self._canvas.mouseLeft.connect(self._on_canvas_mouse_left)
-
-        if self._data_model.data_wrapper is not None:
-            self._fully_synchronize_view()
 
     # -------------- public attributes and methods -------------------------
 
@@ -208,15 +217,22 @@ class ArrayViewer:
 
     def show(self) -> None:
         """Show the viewer."""
-        self._view.set_visible(True)
+        # Queue GUI operation to main thread to ensure thread safety
+        show_future = _app.ndv_app().call_in_main_thread(self._view.set_visible, True)
+        # Block until operation completes to maintain synchronous behavior
+        show_future.result()
 
     def hide(self) -> None:
         """Hide the viewer."""
-        self._view.set_visible(False)
+        # Queue GUI operation to main thread to ensure thread safety
+        hide_future = _app.ndv_app().call_in_main_thread(self._view.set_visible, False)
+        hide_future.result()
 
     def close(self) -> None:
         """Close the viewer."""
-        self._view.set_visible(False)
+        # Queue GUI operation to main thread to ensure thread safety
+        close_future = _app.ndv_app().call_in_main_thread(self._view.set_visible, False)
+        close_future.result()
 
     def clone(self) -> ArrayViewer:
         """Return a new ArrayViewer instance with the same data and display model.


tlambert03:
> My thinking is that in both Python and Java worlds, programmers rarely if ever expect to see a native crash—only exceptions. If we decide not to make show and friends thread-safe in ndv, then my next suggestion is to add a thread check instead, which fails fast with an exception if we aren't on the correct thread, rather than crashing the program outright, which in my view is a bug in the vast majority of circumstances.

#### sys.excepthook

with this phrasing, i just want to double check here that you know about the whole `sys.excepthook` thing and how Qt plays into it.  People are often surprised when they use `QApplication.exec()` to find that python exceptions no longer raise exceptions, but rather crash out to the terminal.  But that's not mandatory... it's due to the fact that all python bindings for Qt (both PySide and PyQt) call `qFatal` during unhandled python exceptions encountered during the event loop.  You can (and most long-lived applications *do*) modify [`sys.excepthook`](https://docs.python.org/3/library/sys.html#sys.excepthook) to prevent crashing out on exceptions.  For example:

- ndv [does it here](https://github.com/pyapp-kit/ndv/blob/35bc88cc6d48aac2ecc9f012683bf974f606c73b/src/ndv/views/_qt/_app.py#L45)
- napari [does it here](https://github.com/napari/napari/blob/0218d54918f842922af3fa5e4e1919ae32164a02/src/napari/utils/notifications.py#L258)
- pymmcore-gui [does it here](https://github.com/pymmcore-plus/pymmcore-gui/blob/db5228fc9491acbc71d0884477f2bef9e18ac29b/src/pymmcore_gui/_app.py#L175-L176)

so, if you're going to "own" the main event loop, and if you're going to commit to using Qt's event loop, then fiji.py should also probably install a custom except hook.

#### app.exec()

I think it's also worth discussing the call to `app.exec()` [here in `fiji.py`](https://github.com/fiji/fiji/blob/05d97b69a1bb90a257b9affa9de278f411a8ded5/config/jaunch/fiji.py#L164)

That *might* be a bit too presumptuous if you want to more generally support python in Fiji.  There are *many* different event loop models in python, plenty of event-loop requiring python modules don't use Qt.  (the main alternates being asyncio/anyio/trio and other structured concurrency libs).

If you call `app.exec()` then i think we _are_ basically stuck running all application logic in another thread, and all downstream libraries will likely need to implement some sort of "call-on-main-thread" workaround.  But you could also have your own concept of an event-loop, similar to how IPython does.  For example, the way IPython supports Qt-based apps (i.e. `%gui qt`) is *not* to put them on another thread; instead, they never actually start a blocking Qt app with `app.exec()`, but rather make clever use of `QtCore.QEventLoop` and `QSocketNotifier`(nix)/`QTimer`(win) and to run both user code and their own prompt on the main thread.  They pump the event loop manually and exit back out repeatedly on user input in the REPL.  [relevant code here](https://github.com/ipython/ipython/blob/main/IPython/terminal/pt_inputhooks/qt.py)... making sure that everything runs on the main thread.

So... Fiji might also try having some magic method for picking an event loop and going back and forth? 

... but it's possibly that AWT/Swing will make this situation harder here than it is over in IPython...

I tried playing around a bit briefly with alternatives to calling `app.exec()` in `fiji.py` (e.g., by polling `app.processEvents()`), but I found that I was unable to get the Java GUI to ever show.  I was a bit surprised to find that without entering `app.exec()`, this call to [`org.scijava.launcher.Splash.show()`](https://github.com/fiji/fiji/blob/05d97b69a1bb90a257b9affa9de278f411a8ded5/config/jaunch/fiji.py#L95-L96) blocked indefinitely... and I guess here now I'm seeing some of the expectations of the native macOS event loop that you were talking about last week.   However, it looks like `QtCore.QEventLoop.exec()`  *does* (via [`processEvents()` inside of `exec()`](https://github.com/qt/qtbase/blob/dev/src/corelib/kernel/qeventloop.cpp#L185-L187) -> call to [`eventDispatcher.loadRelaxed()->processEvents`](https://github.com/qt/qtbase/blob/dev/src/corelib/kernel/qeventloop.cpp#L98-L105) -> call to [`QCocoaEventDispatcher::processEvents`](https://github.com/qt/qtbase/blob/93f59a641e8992cfbdf1ec82cd91cb537dd0d63b/src/plugins/platforms/cocoa/qcocoaeventdispatcher.mm#L286) ... which I think calls NSApp event loops stuff around [here](https://github.com/qt/qtbase/blob/dev/src/plugins/platforms/cocoa/qcocoaeventdispatcher.mm#L347-L359)

I will say that Qt apps are noticeably slower/laggier on IPython `%gui qt` than the standard event loop.  so it's not "free"... but it might be a mode to consider.

anyway, just some thoughts.  will keep playing when I have a moment!

ctrueden:
Thanks @tlambert03, I didn't know about `sys.excepthook` and Qt crashing out the process on exception! That's useful.

I tried adding a `sys.excepthook` handler in `fiji.py`:
```python
# Install custom exception handler to prevent Qt crashes from
# unhandled Python exceptions. This is important because Qt's
# default behavior is to call qFatal and crash the app.
original_excepthook = sys.excepthook

def fiji_excepthook(exc_type, exc_value, exc_traceback):
    # Log the exception.
    _logger.error(
        "Unhandled exception:",
        exc_info=(exc_type, exc_value, exc_traceback),
    )
    # Call original handler (prints to stderr).
    original_excepthook(exc_type, exc_value, exc_traceback)
sys.excepthook = fiji_excepthook
```

but it still crashes when Qt operations happen on the wrong thread.

<details><summary>uncaught NSException crash</summary>

```
WARNING: QObject::installEventFilter(): Cannot filter events for objects in a different thread.
[WARNING:vispy] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'NSWindow should only be instantiated on the main thread!'
*** First throw call stack:
(
	0   CoreFoundation                      0x00000001912a0770 __exceptionPreprocess + 176
	1   libobjc.A.dylib                     0x0000000190d7e418 objc_exception_throw + 88
	2   CoreFoundation                      0x00000001912bc2c0 _CFBundleGetValueForInfoKey + 0
	3   AppKit                              0x00000001962c58d4 -[NSWindow _initContent:styleMask:backing:defer:contentView:] + 260
	4   AppKit                              0x00000001962c5cdc -[NSWindow initWithContentRect:styleMask:backing:defer:] + 48
	5   AppKit                              0x00000001962c5c90 -[NSWindow initWithContentRect:styleMask:backing:defer:screen:] + 24
	6   libqcocoa.dylib                     0x000000010e9cf508 _ZN20QCocoaSystemTrayIcon13emitActivatedEv + 236776
	7   libqcocoa.dylib                     0x000000010e9a52bc _ZN20QCocoaSystemTrayIcon13emitActivatedEv + 64156
	8   libqcocoa.dylib                     0x000000010e99d3a4 _ZN20QCocoaSystemTrayIcon13emitActivatedEv + 31620
	9   libqcocoa.dylib                     0x000000010e99cdb4 _ZN20QCocoaSystemTrayIcon13emitActivatedEv + 30100
	10  QtGui                               0x000000010e2404f4 _ZN14QWindowPrivate6createEb + 464
	11  QtWidgets                           0x000000010ef76144 _ZN14QWidgetPrivate6createEv + 1028
	12  QtWidgets                           0x000000010ef735f4 _ZN7QWidget6createEybb + 376
	13  QtWidgets                           0x000000010f0ec964 _ZNK8QMenuBar15initStyleOptionEP20QStyleOptionMenuItemPK7QAction + 1704
	14  QtWidgets                           0x000000010f0ec55c _ZNK8QMenuBar15initStyleOptionEP20QStyleOptionMenuItemPK7QAction + 672
	15  QtWidgets                           0x000000010f0ecdc4 _ZN8QMenuBarC1EP7QWidget + 216
	16  QtWidgets                           0x000000010f0ad6e8 _ZNK11QMainWindow7menuBarEv + 68
	17  QtWidgets.abi3.so                   0x000000010db48538 _ZL24meth_QMainWindow_menuBarP7_objectS0_ + 88
	18  libpython3.12.dylib                 0x000000010534f5a0 cfunction_call + 308
	19  libpython3.12.dylib                 0x00000001052f1e90 _PyObject_MakeTpCall + 312
	20  libpython3.12.dylib                 0x000000010541b038 _PyEval_EvalFrameDefault + 19184
	21  libpython3.12.dylib                 0x00000001052f22a0 _PyObject_FastCallDictTstate + 160
	22  libpython3.12.dylib                 0x0000000105377e40 slot_tp_init + 312
	23  libpython3.12.dylib                 0x000000010536ede0 type_call + 468
	24  libpython3.12.dylib                 0x00000001052f1e90 _PyObject_MakeTpCall + 312
	25  libpython3.12.dylib                 0x000000010541b038 _PyEval_EvalFrameDefault + 19184
	26  libpython3.12.dylib                 0x00000001052f22a0 _PyObject_FastCallDictTstate + 160
	27  libpython3.12.dylib                 0x0000000105377e40 slot_tp_init + 312
	28  libpython3.12.dylib                 0x000000010536ede0 type_call + 468
	29  libpython3.12.dylib                 0x00000001052f2c88 _PyObject_Call + 176
	30  libpython3.12.dylib                 0x000000010541c034 _PyEval_EvalFrameDefault + 23276
	31  libpython3.12.dylib                 0x0000000105415644 PyEval_EvalCode + 276
	32  libpython3.12.dylib                 0x00000001054120dc builtin_exec + 1488
	33  libpython3.12.dylib                 0x00000001053501cc cfunction_vectorcall_FASTCALL_KEYWORDS + 144
	34  libpython3.12.dylib                 0x00000001052f2a20 PyObject_Vectorcall + 88
	35  libpython3.12.dylib                 0x000000010541b038 _PyEval_EvalFrameDefault + 19184
	36  libpython3.12.dylib                 0x00000001052f61f4 method_vectorcall + 536
	37  _jpype.cpython-312-darwin.so        0x0000000105b3e170 Java_org_jpype_proxy_JPypeProxy_hostInvoke + 196
	38  ???                                 0x0000000132268a88 0x0 + 5136353928
	39  ???                                 0x0000000132264f70 0x0 + 5136338800
	40  ???                                 0x0000000132265420 0x0 + 5136340000
	41  ???                                 0x0000000132265420 0x0 + 5136340000
	42  ???                                 0x0000000132265420 0x0 + 5136340000
	43  ???                                 0x000000013226562c 0x0 + 5136340524
	44  ???                                 0x000000013226517c 0x0 + 5136339324
	45  ???                                 0x0000000132264f70 0x0 + 5136338800
	46  ???                                 0x0000000132265420 0x0 + 5136340000
	47  ???                                 0x0000000132264f70 0x0 + 5136338800
	48  ???                                 0x0000000132265420 0x0 + 5136340000
	49  ???                                 0x000000013226562c 0x0 + 5136340524
	50  ???                                 0x000000013226517c 0x0 + 5136339324
	51  ???                                 0x000000013226562c 0x0 + 5136340524
	52  ???                                 0x000000013226517c 0x0 + 5136339324
	53  ???                                 0x0000000132260144 0x0 + 5136318788
	54  libjvm.dylib                        0x0000000112fd9444 _ZN9JavaCalls11call_helperEP9JavaValueRK12methodHandleP17JavaCallArgumentsP10JavaThread + 992
	55  libjvm.dylib                        0x0000000112fd8374 _ZN9JavaCalls12call_virtualEP9JavaValueP5KlassP6SymbolS5_P17JavaCallArgumentsP10JavaThread + 320
	56  libjvm.dylib                        0x0000000112fd8440 _ZN9JavaCalls12call_virtualEP9JavaValue6HandleP5KlassP6SymbolS6_P10JavaThread + 100
	57  libjvm.dylib                        0x00000001130aa10c _ZL12thread_entryP10JavaThreadS0_ + 156
	58  libjvm.dylib                        0x0000000112fedb74 _ZN10JavaThread17thread_main_innerEv + 152
	59  libjvm.dylib                        0x0000000113529d60 _ZN6Thread8call_runEv + 200
	60  libjvm.dylib                        0x000000011334b108 _ZL19thread_native_entryP6Thread + 280
	61  libsystem_pthread.dylib             0x00000001911b0c08 _pthread_start + 136
	62  libsystem_pthread.dylib             0x00000001911abba8 thread_start + 8
)
2025-11-13 13:21:30.839 fiji-macos-arm64[91819:13258507] Apple AWT Internal Exception: NSWindow should only be instantiated on the main thread!
2025-11-13 13:21:30.839 fiji-macos-arm64[91819:13258507] trace: (
	0   CoreFoundation                      0x00000001912a0770 __exceptionPreprocess + 176
	1   libobjc.A.dylib                     0x0000000190d7e418 objc_exception_throw + 88
	2   CoreFoundation                      0x00000001912bc2c0 _CFBundleGetValueForInfoKey + 0
	3   AppKit                              0x00000001962c58d4 -[NSWindow _initContent:styleMask:backing:defer:contentView:] + 260
	4   AppKit                              0x00000001962c5cdc -[NSWindow initWithContentRect:styleMask:backing:defer:] + 48
	5   AppKit                              0x00000001962c5c90 -[NSWindow initWithContentRect:styleMask:backing:defer:screen:] + 24
	6   libqcocoa.dylib                     0x000000010e9cf508 _ZN20QCocoaSystemTrayIcon13emitActivatedEv + 236776
	7   libqcocoa.dylib                     0x000000010e9a52bc _ZN20QCocoaSystemTrayIcon13emitActivatedEv + 64156
	8   libqcocoa.dylib                     0x000000010e99d3a4 _ZN20QCocoaSystemTrayIcon13emitActivatedEv + 31620
	9   libqcocoa.dylib                     0x000000010e99cdb4 _ZN20QCocoaSystemTrayIcon13emitActivatedEv + 30100
	10  QtGui                               0x000000010e2404f4 _ZN14QWindowPrivate6createEb + 464
	11  QtWidgets                           0x000000010ef76144 _ZN14QWidgetPrivate6createEv + 1028
	12  QtWidgets                           0x000000010ef735f4 _ZN7QWidget6createEybb + 376
	13  QtWidgets                           0x000000010f0ec964 _ZNK8QMenuBar15initStyleOptionEP20QStyleOptionMenuItemPK7QAction + 1704
	14  QtWidgets                           0x000000010f0ec55c _ZNK8QMenuBar15initStyleOptionEP20QStyleOptionMenuItemPK7QAction + 672
	15  QtWidgets                           0x000000010f0ecdc4 _ZN8QMenuBarC1EP7QWidget + 216
	16  QtWidgets                           0x000000010f0ad6e8 _ZNK11QMainWindow7menuBarEv + 68
	17  QtWidgets.abi3.so                   0x000000010db48538 _ZL24meth_QMainWindow_menuBarP7_objectS0_ + 88
	18  libpython3.12.dylib                 0x000000010534f5a0 cfunction_call + 308
	19  libpython3.12.dylib                 0x00000001052f1e90 _PyObject_MakeTpCall + 312
	20  libpython3.12.dylib                 0x000000010541b038 _PyEval_EvalFrameDefault + 19184
	21  libpython3.12.dylib                 0x00000001052f22a0 _PyObject_FastCallDictTstate + 160
	22  libpython3.12.dylib                 0x0000000105377e40 slot_tp_init + 312
	23  libpython3.12.dylib                 0x000000010536ede0 type_call + 468
	24  libpython3.12.dylib                 0x00000001052f1e90 _PyObject_MakeTpCall + 312
	25  libpython3.12.dylib                 0x000000010541b038 _PyEval_EvalFrameDefault + 19184
	26  libpython3.12.dylib                 0x00000001052f22a0 _PyObject_FastCallDictTstate + 160
	27  libpython3.12.dylib                 0x0000000105377e40 slot_tp_init + 312
	28  libpython3.12.dylib                 0x000000010536ede0 type_call + 468
	29  libpython3.12.dylib                 0x00000001052f2c88 _PyObject_Call + 176
	30  libpython3.12.dylib                 0x000000010541c034 _PyEval_EvalFrameDefault + 23276
	31  libpython3.12.dylib                 0x0000000105415644 PyEval_EvalCode + 276
	32  libpython3.12.dylib                 0x00000001054120dc builtin_exec + 1488
	33  libpython3.12.dylib                 0x00000001053501cc cfunction_vectorcall_FASTCALL_KEYWORDS + 144
	34  libpython3.12.dylib                 0x00000001052f2a20 PyObject_Vectorcall + 88
	35  libpython3.12.dylib                 0x000000010541b038 _PyEval_EvalFrameDefault + 19184
	36  libpython3.12.dylib                 0x00000001052f61f4 method_vectorcall + 536
	37  _jpype.cpython-312-darwin.so        0x0000000105b3e170 Java_org_jpype_proxy_JPypeProxy_hostInvoke + 196
	38  ???                                 0x0000000132268a88 0x0 + 5136353928
	39  ???                                 0x0000000132264f70 0x0 + 5136338800
	40  ???                                 0x0000000132265420 0x0 + 5136340000
	41  ???                                 0x0000000132265420 0x0 + 5136340000
	42  ???                                 0x0000000132265420 0x0 + 5136340000
	43  ???                                 0x000000013226562c 0x0 + 5136340524
	44  ???                                 0x000000013226517c 0x0 + 5136339324
	45  ???                                 0x0000000132264f70 0x0 + 5136338800
	46  ???                                 0x0000000132265420 0x0 + 5136340000
	47  ???                                 0x0000000132264f70 0x0 + 5136338800
	48  ???                                 0x0000000132265420 0x0 + 5136340000
	49  ???                                 0x000000013226562c 0x0 + 5136340524
	50  ???                                 0x000000013226517c 0x0 + 5136339324
	51  ???                                 0x000000013226562c 0x0 + 5136340524
	52  ???                                 0x000000013226517c 0x0 + 5136339324
	53  ???                                 0x0000000132260144 0x0 + 5136318788
	54  libjvm.dylib                        0x0000000112fd9444 _ZN9JavaCalls11call_helperEP9JavaValueRK12methodHandleP17JavaCallArgumentsP10JavaThread + 992
	55  libjvm.dylib                        0x0000000112fd8374 _ZN9JavaCalls12call_virtualEP9JavaValueP5KlassP6SymbolS5_P17JavaCallArgumentsP10JavaThread + 320
	56  libjvm.dylib                        0x0000000112fd8440 _ZN9JavaCalls12call_virtualEP9JavaValue6HandleP5KlassP6SymbolS6_P10JavaThread + 100
	57  libjvm.dylib                        0x00000001130aa10c _ZL12thread_entryP10JavaThreadS0_ + 156
	58  libjvm.dylib                        0x0000000112fedb74 _ZN10JavaThread17thread_main_innerEv + 152
	59  libjvm.dylib                        0x0000000113529d60 _ZN6Thread8call_runEv + 200
	60  libjvm.dylib                        0x000000011334b108 _ZL19thread_native_entryP6Thread + 280
	61  libsystem_pthread.dylib             0x00000001911b0c08 _pthread_start + 136
	62  libsystem_pthread.dylib             0x00000001911abba8 thread_start + 8
)
libc++abi: terminating due to uncaught exception of type NSException
./fiji: line 55: 91819 Abort trap: 6           "$launcher" "$@"
```

</details>

Here is what Claude said about it:

> The issue is that `sys.excepthook` only handles unhandled exceptions that propagate to the top level. It doesn't catch exceptions that occur inside Qt callbacks or other frameworks' event loops.
> 
> The crash you're seeing is happening because:
> 1. Qt widget creation (QMenuBar/QMainWindow) is being called from a non-main thread (the Fiji thread)
> 2. This triggers an NSException at the Objective-C level on macOS
> 3. NSException is not a Python exception, so `sys.excepthook` never sees it
> 4. The app crashes before Python's exception handling even gets involved
> 
> The real problem is that your architecture has Fiji running in a background thread while Qt runs on the main thread, but Java/Swing code is trying to create Qt widgets from the Fiji thread. This violates the macOS/Qt threading requirement.
> 
> **The `sys.excepthook` approach won't help here** because:
> - The exception is happening at the C/Objective-C level, not Python
> - It's a threading violation, not a recoverable Python exception
> - Even if caught, you can't "fix" the threading issue after the fact
> 
> **Solutions:**
> 
> 1. **Best approach:** Keep using @ensure_main_thread decorator where needed - this is the correct solution. It ensures Qt operations happen on the main thread.
> 2. **If you want broader protection**, you could patch Qt object creation to automatically dispatch to main thread, but this is complex and fragile.
> 3. **sys.excepthook is still useful** for other cases (like pure Python exceptions in Qt slots), so keep it, but understand its limitations: it can't save you from threading violations at the native level.

Pushing Claude a bit more, we tried some other approaches:

1.  Using PyObjC or ctypes, install an **Objective-C exception handler to catch NSExceptions** (like threading violations) that would otherwise crash.

    ❌ Still crashes, due to internal exception handling in Java AWT:
    
    > The handler is set up correctly, but it's not being called. The key issue is in the stack trace:
    > 
    > ```
    > 2025-11-13 13:38:59.170 fiji-macos-arm64[93160:13275157] Apple AWT Internal Exception: NSWindow should only be instantiated on the main thread!
    > ```
    > 
    > "Apple AWT Internal Exception" - this means the Java AWT layer is catching the NSException first and handling it internally, preventing it from reaching your NSSetUncaughtExceptionHandler.
    > 
    > The problem is that the exception is being caught and re-thrown by Java's AWT code, which then calls abort() directly, bypassing the normal exception handling mechanism.

2. Patch `threading.Thread.run` to catch Python exceptions on any thread.
    ```python
    # Patch threading.Thread.run to catch all exceptions in threads,
    # including those from Java->Python callbacks via JPype.
    original_thread_run = threading.Thread.run
    
    def patched_thread_run(self):
        try:
            original_thread_run(self)
        except Exception:
            # Log the exception but don't let it crash the thread.
            _logger.error(
                f"Exception in thread {self.name}:",
                exc_info=True,
            )
            # Also print to stderr for visibility.
            traceback.print_exc()
    ```

    ❌ Still crashes. The crash comes from the JVM, not from a Python thread.

3.  Try **noticing SIGABRT by registering a callback via `signal.signal`**:
    ```python
    # Define our signal handler.
    SIGHANDLER = ctypes.CFUNCTYPE(None, ctypes.c_int)
    
    def sigabrt_handler(signum):
        """Handle SIGABRT without crashing."""
        _logger.error("Caught SIGABRT - preventing crash")
        print("*** Caught SIGABRT (NSException threading violation)", file=sys.stderr)
        # Don't exit - just return and hope for the best.
    
    _sigabrt_callback = SIGHANDLER(sigabrt_handler)
    
    # Install the handler using signal.signal().
    # This is simpler than sigaction and should work.
    signal.signal(signal.SIGABRT, lambda sig, frame: sigabrt_handler(sig))
    ```

    :x: Still crashes, because:

    > The SIGABRT handler isn't being called because the exception is being thrown inside a Java thread (see the stack trace - it's in libjvm.dylib), and Python signal handlers only work on the main Python thread.
    >
    > Since the crash is happening because abort() is called by the C++ runtime (libc++abi: terminating due to uncaught exception), we need to intercept it at a lower level. The fundamental problem is that we can't easily catch NSExceptions thrown from JVM threads.

4.  Install a C++ terminate handler using `std::set_terminate`. Unfortunately, the symbol is mangled on macOS, so the code looks like this:
    ```python
    # The symbol is mangled as _ZSt13set_terminatePFvvE on macOS/Linux.
    libcxx._ZSt13set_terminatePFvvE.restype = ctypes.c_void_p
    libcxx._ZSt13set_terminatePFvvE.argtypes = [TERMINATE_HANDLER]
    libcxx._ZSt13set_terminatePFvvE(_cpp_terminate_callback)
    _logger.debug("Installed C++ terminate handler")
    ```

    ✅ WORKS! The complete function is 118 lines of Python code, but it does the trick.

Here's what I ended up with:

https://github.com/fiji/fiji/blob/f1460d8d01d092fe93d0ffc325db469c155af538/config/fiji.py

Still crashes when you forget the `@ensure_main_thread`, but not before explaining *why* it's crashing, and how to avoid it next time. 🤷

marktsuchida:
Very interesting -- I didn't know that the C++ terminate handler gets called for ObjC exceptions (I guess it makes sense given the shared Itanium ABI).

`std::set_terminate()` has the usual problem of "what happens if somebody else also changes the (global) handler?", but of course if anybody should, it should be the app, not a library, so doing it in Fiji makes sense.

One thing to keep in mind is that `std::terminate()` is called in several different situations (not just unhandled exceptions) in C++ (exception thrown while unwinding the stack, `noexcept` function threw exception, exception escaped thread, joinable thread destroyed, mutex destroyed while locked, terminate handler threw, pure virtual function called, etc.).

The default terminate handler (depending on the platform, but definitely on macOS) prints some useful information, like the type and message of the uncaught exception, if that was the cause, or whatever else the cause was. So it might be good to call the default terminate handler instead of (or before) just exiting (and btw `sys.exit()` seems dangerous because it may run atexit handlers among other things).

`std::set_terminate()` returns a function pointer to the previous terminate handler, so it can be stored and called. (And it looks like the default handler already prints the type and message for ObjC exceptions, and even adds a stack trace.)

I'd also note that the ObjC `NSInternalInconsistencyException` is used for non-threading-related issues, too. An example I remember from many years ago is if the data source for a UI table gets the math wrong. Not to mention generic `NSException` which could be a lot of things (somewhat analogous to `java.lang.Error`).

So it might be nice to print a hint that GUI threading could be a possible cause, but maybe not sound too definite, lest it mislead the reader.

I also don't know what the chances are of the Python interpreter being in a consistent state when the terminate handler is called. If `std::terminate()` is called on a C++ thread that never acquired the GIL, while another thread is hung or crashed holding the GIL, I believe it will deadlock. Even if `std::terminate()` is called on the same thread that currently holds the GIL, it will likely find the interpreter in the wrong state (interpreter has called out into C while the terminate handler tries to run Python code without having returned from that C call).

Probably best if production terminate handlers are implemented in C or C++ and kept minimal so that nothing futher can go wrong.

Issue #275, opened 2026-05-03
multiscale chunked rendering


kevinyamauchi:
## Overview

I think it would be nice to be able to view larger-than-memory images in Python.  I have a multiscale rendering prototype working with `pygfx` and as I saw multi-resolution data on [the roadmap](https://pyapp-kit.github.io/ndv/latest/motivation/?h=roadma#scope-and-roadmap), I was wondering if `ndv` could be a good home for it. I'm sorry, it's a bit of a brain dump. Happy to jump on a call if you'd like a walkthrough.

## Current prototype

I have made a prototype of 2d and 3d multiscaled chunk rendering in pygfx (https://github.com/kevinyamauchi/demo-multiscale). The basic concept is that when the view is updated (i.e., camera moves or dims state changes), only the data that is both in view and fits in the GPU is loaded. As we have multiple resolutions of the data, we can select data that has a resolution such that the level of detail matches that of the screen. I've given a brief overview below of how it works.

**Data structure and IO**
Currently, I am using tensorstore to load ome-zarr v0.5 files. It should be straightforward to make adapters for other formats and IO libraries. In ndv, I believe this would be a `DataWrapper`

**Selecting data in view**
We can use the camera model to determine which region in the displayed world space is viewable by the camera (e.g., the camera frustum in 3d). Then we can select all of the chunks that are in view.  Currently I take any chunk that has a corner in the frustum, but there are other ways to do it. 

**Selecting which resolution to load from**
In 2d, the level of detail is selected for the full slice to match the screen pixel size to the data pixel size. For 3d selection is done on a per-brick level, selecting the scale to match the brick size to the screen pixel. For both 2d and 3d, there is also an `lod_bias` parameter that can tune the bias towards finer or coarser levels of detail.

**Texture indirection**
One of the performance bottlenecks is loading the data onto the GPU. As you man imagine when the camera moves, we can often reuse some of the chunks as they remain in view.  We can use a cache on the GPU to re-use data and save both the time to load the chunk from disk and the time to load the chunk to the GPU. Rather than moving the data from the cache to the sampling texture, we can use a technique called "texture indirection" point the shader sampler to the right element in the cache.  [This writeup](https://dev.to/mpanknin/kiln-webgpu-native-out-of-core-volume-rendering-for-multi-gb-datasets-2alb) on Kiln has a nice explanation of how texture indirection can be used in volume rendering (works similarly in 2d). The implementation in my prototype has a similar design to Kiln.

[...]

PR #276, opened 2026-05-04
WIP: use scenex

gselzer:
# Overview

This PR aims to replace the references to multiple canvas backends (vispy, pygfx) with a single implementation against [scenex](https://github.com/pyapp-kit/scenex) models. It's a large change, deserving lots of review and scrutiny.

# Reasoning

1) **One canvas implementation** The flexibility ndv offers users presents significant effort for ndv developers who want to enable a feature across both canvas implementations. Both `src/ndv/views/_vispy` and `src/ndv/views/_pygfx` have thousands of lines of code and separate test bases. As this PR shows, adopting scenex results in far fewer raw lines of code to maintain.

2) **Robust scene models** Adding more functionality (including the ability to display multiple datasets, multiscale data as suggested in #275, orthoviews, etc.) will require knowledge of the scenegraph state. There are already bits and pieces of this in ndv (e.g. `ImageHandle`, `RectangularROIHandle`), but scenex exists for this purpose.

# Implementation Goals

The primary goal of this PR is to condense down to a single canvas implementation, removing dependencies on vispy and pygfx in favor of dependencies of scenex, while maintaining the same public API, and all (excluding tests aimed at removed behavior) tests.

@tlambert03 your review is always welcome, but might be good to wait a little longer, as I fear that despite passing tests on my machine this PR will show me that there are many more things to clean up!


tests/test_shared_histogram.py
```diff
@@ -377,39 +346,291 @@ def test_opening_histogram_in_grayscale_respects_mode() -> None:
...
-    mock_hist = ctrl._shared_histogram
 
-    ctrl._highlight_values({}, (5, 5))
-    mock_hist.highlight.assert_called_with({})
+    with patch.object(ctrl._shared_histogram, "highlight") as mock_highlight:
+        ctrl._highlight_values({}, (5, 5))
+    mock_highlight.assert_called_with({})
+
+
+# ---------- Visual behavior tests ----------
```
gselzer, line 389: Noting that these tests were moved over from `tests/views/_vispy/test_shared_histogram.py`

tests/views/_vispy/test_shared_histogram.py
```diff
@@ -1,279 +0,0 @@
-"""Tests for VispySharedHistogramCanvas visual behavior."""
```
gselzer, line 1: These tests were moved to `tests/test_shared_histogram.py`

tests/views/_vispy/test_volume_downsample.py
```diff
@@ -1,168 +0,0 @@
-from __future__ import annotations
```
gselzer, line 1: This is really a scenex concern, see how these things are tested e.g. [here](https://github.com/pyapp-kit/scenex/blob/6bed52c651a92c5aaea9a4a6f5ca2a8f5338f6a2/tests/adaptors/_vispy/test_volume.py#L118)

tests/views/_vispy/test_histogram.py
```diff
@@ -1,163 +0,0 @@
-from __future__ import annotations
```
gselzer, line 1: These tests were moved to `test/test_histogram.py`

src/ndv/views/bases/_app.py
gselzer, line 30: Note that it would be nice to just get rid of this class, however it's public API

src/ndv/views/bases/__init__.py
```diff
@@ -4,23 +4,13 @@
...
 from ._app import NDVApp
 from ._array_view import ArrayView
-from ._graphics._canvas import ArrayCanvas, HistogramCanvas, SharedHistogramCanvas
-from ._graphics._canvas_elements import CanvasElement, ImageHandle, RectangularROIHandle
-from ._graphics._mouseable import Mouseable
 from ._lut_view import LUTView
 from ._view_base import Viewable
 
 __all__ = [
-    "ArrayCanvas",
```
gselzer, line 14: Just noting that there's lots of public API here being deleted already

src/ndv/views/__init__.py
```diff
@@ -19,12 +16,9 @@
 )
 
 __all__ = [
-    "CanvasBackend",
     "GuiFrontend",
     "call_later",
-    "get_array_canvas_class",
     "get_array_view_class",
```
gselzer, line 21: More public API being deleted

src/ndv/views/_app.py
gselzer, line 1: This is another file ripe for pruning.

gselzer:
@tlambert03 I think I'm getting happier with how this is looking, and would be interested in your initial thoughts here. As I mentioned above, this is a rather large amount of changing code, and it would need a release of scenex before merge. More than anything else, I'm interested in your thoughts about the effects to the project structure, including:
* The removal of `src/ndv/views/_vispy` and `src/ndv/views/_pygfx`, in favor of scenex models in `src/ndv/views`
* The removal of `src/ndv/views/bases/_graphics`, and the slimming down of `src/ndv/views/bases`.
* The slimming down of `src/ndv/views/_qt` et al - there's actually probably more we could trim here if we wanted.
* The consolidation of several tests.

Despite all of that removal, I think all ndv features are fully functional (let me know if any testing you do uncovers anything I missed).

